PR #1, opened 2017-12-29
Path stuff

yoptar:


deeppavlov/core/commands/utils.py
```diff
@@ -9,19 +9,17 @@ def set_usr_dir(config_path: str, usr_dir_name='USR_DIR') -> PurePath:
     Make a serialization user dir.
     """
     config = read_json(config_path)
-    try:
```
my-master: try...except KeyError is more pythonic way to deal with dictionary keys checking than if...else

deeppavlov/core/commands/utils.py
```diff
@@ -9,19 +9,17 @@ def set_usr_dir(config_path: str, usr_dir_name='USR_DIR') -> PurePath:
...
+    if 'usr_dir' in config:
         usr_dir = Path(config['usr_dir'])
-    except KeyError:
-        parent = Path(config_path).resolve().parent
-        usr_dir = parent.joinpath(usr_dir_name)
+    else:
+        usr_dir = Path(config_path).expanduser().absolute().parent / usr_dir_name
```
my-master: Currently usr_dir should be defined relatively to config path, easier to perform experiments and debugging.

deeppavlov/core/commands/utils.py
```diff
@@ -9,19 +9,17 @@ def set_usr_dir(config_path: str, usr_dir_name='USR_DIR') -> PurePath:
...
 
-    if not usr_dir.exists():
-        usr_dir.mkdir()
+    usr_dir.mkdir(mode=0o755, exist_ok=True)
```
my-master: Better: usr_dir.mkdir(exist_ok=True)

deeppavlov/core/commands/utils.py
```diff
@@ -9,19 +9,17 @@ def set_usr_dir(config_path: str, usr_dir_name='USR_DIR') -> PurePath:
...
 
     paths.USR_PATH = usr_dir
     return usr_dir
 
 
 def set_vocab_path() -> PurePath:
-    return paths.USR_PATH.joinpath('vocab.txt')
+    return paths.USR_PATH / 'vocab.txt'
```
my-master: Ok

deeppavlov/core/commands/utils.py
yoptar: default value for mode is `0o777`, so i think mode should not be omitted

my-master: Default umask is applied to a newly created dir via mkdir(). So when the mode is 0o777, the actual permission of the dir is 0755 (if umask is 022).

my-master: System umask applies to mkdir() default value, so if the parent dir has 0o755 permission mode, the created dir won't have 0o777 permissions even if default mkdir() mode is 0o777. Here is more about the subject:
https://bugs.python.org/issue5220

my-master:
Pull -request merged.

PR #2, opened 2018-01-11
Added pretrained-vectors.md with trained embeddings download links

litinsky:
Added pretrained-vectors.md with trained embeddings download links.


Timopheym:
Hey, don't you whant add http://rusvectores.org/ru/about/ too? Is it supported out of the box?

PR #4, opened 2018-01-17
feat: add wraps

vikmary:
- [x] preserve docs while wrapping functions


my-master:
Ok, but this is not a "feat" because no new functionality was added. This is a "fix".

PR #7, opened 2018-01-23
Classification

dilyararimovna:
add vocabs, tokenizer, readme, docstrings, fixed infer and training of kerasintentmodel


deeppavlov/core/data/dataset.py
```diff
@@ -34,24 +51,35 @@ def __init__(self, data: Dict[str, List[Tuple[Any, Any]]], seed: int = None, *ar
...
         Args:
             batch_size (int): number of samples in batch
             data_type (str): can be either 'train', 'test', or 'valid'
+            shuffle (bool): whether to shuffle dataset before batching
+            seed (int): random seed for batching
         Returns:
             batch_gen (Generator): a generator, that iterates through the part (defined by data_type) of the dataset
         """
+        if shuffle is not None:
+            self.shuffle = shuffle
```
yoptar: I don't think that batch generator should be able to switch the shuffle flag for the dataset forever

deeppavlov/core/data/dataset.py
```diff
@@ -34,24 +51,35 @@ def __init__(self, data: Dict[str, List[Tuple[Any, Any]]], seed: int = None, *ar
...
         order = list(range(data_len))
-
-        rs = random.getstate()
-        random.setstate(self.random_state)
-        random.shuffle(order)
-        self.random_state = random.getstate()
-        random.setstate(rs)
+        if self.shuffle:
+            rs = random.getstate()
+            random.seed(self.seed)
```
yoptar: Every epoch will most likely have the same seed, so the shuffling is always the same so there is basically no shuffling

PR #61, opened 2018-02-19
Feature/new tf model

mu-arkhipov:
TFModel refactored. NER, SlotFIll, and GoalOrientedBotNetwork rewritten to support new class.
What is done:
- TFModel refactored. Now it is more like regular tf model
- saving and restoring trainable variables (optimizer variables are excluded from saving and restoration)
- getting train_op become easy. batch_norm updates and gradients clipping are supported from the box 

Attention! To use load/save provided in TFModel the session must be attribute of the model with name `self.sess`!

### DIFF
--- a/deeppavlov/core/models/tf_model.py
+++ b/deeppavlov/core/models/tf_model.py
@@ -13,119 +13,112 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
-
 import sys
 from abc import abstractmethod
+from collections import defaultdict
+import numpy as np
 
 import tensorflow as tf
-from overrides import overrides
 
 from deeppavlov.core.models.nn_model import NNModel
-from deeppavlov.core.models.serializable import Serializable
-from deeppavlov.core.common.attributes import check_attr_true
 from deeppavlov.core.common.log import get_logger
-from deeppavlov.core.common.errors import ConfigError
 from .tf_backend import TfModelMeta
-
-"""
-Here is an abstract class for neural network models based on Tensorflow.
-If you use something different, ex. Pytorch, then write similar to this class, inherit it from
-Trainable and Inferable interfaces and make a pull-request to deeppavlov.
-"""
 
 log = get_logger(__name__)
 
 
-class TFModel(Serializable, metaclass=TfModelMeta):
-    def __init__(self, **kwargs):
-        self._saver = tf.train.Saver
-        super().__init__(**kwargs)
-
-    @abstractmethod
-    def _add_placeholders(self):
-        """
-        Add all needed placeholders for a computational graph.
-        """
-        pass
-
-    @abstractmethod
-    def run_sess(self, *args, **kwargs):
-        """
-        1. Call _build_graph()
-        2. Define all comuptations.
-        3. Run tf.sess.
-        3. Reset state if needed.
-        :return:
-        """
-        pass
-
-    @abstractmethod
-    def _train_step(self, features, *args):
-        """
-        Define a single training step. Feed dict to tf session.
-        :param features: input features
-        :param args: any other inputs, including target vector, you need to pass for training
-        :return: metric to return, usually loss
-        """
-        pass
-
-    @abstractmethod
-    def _forward(self, features, *args):
-        """
-        Pass an instance to get a prediction.
-        :param features: input features
-        :param args: any other inputs you need to pass for training
-        :return: prediction
-        """
-        pass
-
-    @check_attr_true('train_now')
-    def train(self, features, *args, **kwargs):
-        """
-        Just a wrapper for a private method.
-        """
-        return self._train_step(features, *args, **kwargs)
-
-    def infer(self, instance, *args, **kwargs):
-        """
-        Just a wrapper for a private method.
-        """
-        return self._forward(instance, *args, **kwargs)
-
-    def save(self):
-        save_path = str(self.save_path)
-        saver = tf.train.Saver()
-        log.info('[saving model to {}]'.format(save_path))
-        saver.save(self.sess, save_path)
-        log.info('model saved')
-
-    def get_checkpoint_state(self):
-        if self.load_path:
-            try:
-                if self.load_path.parent.is_dir():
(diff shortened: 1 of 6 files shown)


deeppavlov/core/models/tf_model.py
```diff
@@ -13,119 +13,123 @@
...
-        Define a single training step. Feed dict to tf session.
-        :param features: input features
-        :param args: any other inputs, including target vector, you need to pass for training
-        :return: metric to return, usually loss
         """
         pass
 
+    def load(self):
+        """Load model parameters from self.load_path"""
+        path = str(self.load_path.resolve())
```
yoptar: Why would you need to resolve save and load paths here? Are they not resolved in the Serializable class?

mu-arkhipov: save/load pathes are user_expanded but not resolved:
```def expand_path(path: Union[str, Path]) -> Path:
    return get_deeppavlov_root() / Path(path).expanduser()```

File: deeppavlov/core/models/tf_model.py [hunk 2]
```diff
@@ -13,119 +13,123 @@
...
+            # Exclude optimizer variables from saved variables
+            var_list = [var for var in tf.trainable_variables()
+                        if not var.name.startswith('Optimizer')]
+            saver = tf.train.Saver(var_list)
+            saver.restore(self.sess, path)
+
+    def save(self):
+        """Save model parameters to self.save_path"""
+        path = str(self.save_path.resolve())
+        print('[saving model to {}]'.format(path), file=sys.stderr)
```
yoptar: Use logging instead of prints, please.

deeppavlov/core/models/tf_model.py
```diff
@@ -13,119 +13,123 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
 
     @abstractmethod
-    def _add_placeholders(self):
-        """
-        Add all needed placeholders for a computational graph.
-        """
-        pass
+    def train_on_batch(self, x_batch, y_batch):
```
yoptar: This abstract method is defined in the parent class NNModel, so it's not needed here.

mu-arkhipov: The same thing with __call__, am I right?

yoptar: Yes

deeppavlov/models/ner/ner.py
```diff
@@ -50,14 +44,11 @@ def __init__(self, **kwargs):
...
-        train_parameters = {par: opt[par] for par in train_parameters_names if par in opt}
+        train_parameters = {par: self.opt[par] for par in train_parameters_names if par in self.opt}
         self.train_parameters = train_parameters
 
-        self.opt = opt
-
         # Try to load the model (if there are some model files the model will be loaded from them)
-        super().__init__(save_path=save_path, load_path=load_path,
-                         train_now=train_now, mode=mode)
+        super().__init__(**self.opt)
```
yoptar: Why do you send self.opt to super() and not kwargs? Also, in slotfill.py it's different

deeppavlov/core/models/tf_model.py
```diff
@@ -13,119 +13,97 @@
...
-        """
-        return self._forward(instance, *args, **kwargs)
+    def load(self):
+        """Load model parameters from self.load_path"""
+        path = str(self.load_path.resolve())
+        # Check presence of the model files
+        if tf.train.checkpoint_exists(path):
+            log.info('[loading model from {}]'.format(path))
+            # Exclude optimizer variables from saved variables
+            var_list = [var for var in tf.trainable_variables()
```
vikmary: Maybe we should make variables' exclusion optional (via method parameter?)
Or
```python
def load(self, exclude_scopes=['Optimizer']):
    <...>
    var_list = self.get_trainable_variables(exclude_scopes)
    <..>
def _get_trainable_variables(scope_names=[]):
       all_vars = tf.trainable_variables()
       if scope_names:
            vars_to_train = set()
            for sn in scope_names:
                scope_vars = tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES, scope=sn)
                vars_to_train.update(scope_vars)
            return list(vars_to_train)
        else:
            return all_vars
```

mu-arkhipov:
```
def load(self, exclude_scopes=['Optimizer']):
    <...>
    var_list = self.get_trainable_variables(exclude_scopes)
    <..>
def _get_trainable_variables(exclude_scopes=[]):
       all_vars = tf.trainable_variables()
       vars_to_train = [var for var in all_vars if all(sc not in var.name for sc in exclude_scopes)]
       return vars_to_train
```

deeppavlov/core/models/tf_model.py
```diff
@@ -13,119 +13,97 @@
...
         """
-        Load session from checkpoint
+        with tf.variable_scope('Optimizer'):
+            if learnable_scopes is None:
+                variables_to_train = tf.trainable_variables()
+            else:
+                variables_to_train = []
+                for scope_name in learnable_scopes:
+                    for var in tf.trainable_variables():
+                        if var.name.startswith(scope_name):
```
vikmary: `var.name.startswith(scope_name)` -> `scope_name not in var.name`?

deeppavlov/models/ner/network.py
```diff
@@ -59,7 +54,8 @@ def __init__(self,
                  net_type='cnn',
                  char_filter_width=5,
                  verbouse=False,
```
vikmary: verbouse -> verbose?=)

deeppavlov/core/models/tf_model.py
mu-arkhipov: done

deeppavlov/core/models/tf_model.py
mu-arkhipov: in progress

deeppavlov/models/ner/network.py
mu-arkhipov: fixed

deeppavlov/core/models/tf_model.py
mu-arkhipov: Done

PR #91, opened 2018-03-07
test(): add test for interacting with custom queries

nikolay-bushkov:
After refactoring, it is possible to easily add list of query-response
pairs for every model (config), which will be used to compare pretrained
model output with expected output. Initial lists added for error_model
and ner. Also URL for downloading pretrained ner_conll2003_model added
IP-1344 #done

### DIFF
--- a/tests/test_quick_start.py
+++ b/tests/test_quick_start.py
@@ -1,40 +1,70 @@
 import pytest
 from pathlib import Path
+import pexpect
 import json
-import subprocess as sp
 import shutil
 
 
 tests_dir = Path(__file__, '..').resolve()
 
-# Mapping from model names to configs and corresponding Query-Response pairs
-MCQR = {"error_model": {"configs/error_model/brillmoore_wikitypos_en.json": ("error_model", "", ""),
-                        "configs/error_model/brillmoore_kartaslov_ru.json": ("error_model", "", "")
-                        },
-        "go_bot": {"configs/go_bot/gobot_dstc2.json": ("go_bot", "", ""),
-                   # "configs/go_bot/config_all.json":
-                   #     ("go_bot_all", "", ""),
-                   "configs/go_bot/gobot_dstc2_minimal.json": ("go_bot_minimal", "", "")
-                   },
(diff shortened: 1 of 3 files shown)


tests/test_quick_start.py
```diff
@@ -46,42 +76,45 @@ def setup_module():
...
-        if not tests_dir.parent.joinpath('download').exists():
+    def interact(conf_file, model_dir, qr_list=None):
+        qr_list = qr_list or []
+        p = pexpect.spawn("python3", ["-m", "deeppavlov.deep", "interact", str(conf_file)], timeout=None)
+        for (query, expected_response) in qr_list:  # works until the first failed query
+            p.expect(":: ")
+            p.sendline(query)
+            p.expect(">> ")
+            actual_response = p.readline().decode().strip()
+            assert expected_response == actual_response, f"Error in interacting with {model_dir} ({conf_file}): {query}"
```
yoptar: It would be more informative to also show expected and actual responses in the error message.
Also, is it possible to run all the queries and not stop after the first failure?

tests/test_quick_start.py
```diff
@@ -1,40 +1,70 @@
...
+                          # ("Albert Einstein and Erwin Schrodinger", "['B-PER', 'I-PER', 'O', 'B-PER', 'I-PER']"),
+                          # ("Antananarivo is the capital of Madagascar", "['B-LOC', 'O', 'O', 'O', 'O', 'B-LOC']"),
+                          # ("UN launches new global data collection tool to help reduce disaster",
+                          #  "['B-ORG', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O']")
+                      ],
+                  ("configs/ner/ner_dstc2.json", "ner"):
+                      [
+                          # ("chinese food", "['B-food', 'O']"),
+                          # ("in the west part", "['O', 'O', 'B-area', 'O']"),
+                          # ("moderate price range", "['B-pricerange', 'O', 'O']")
```
yoptar: Why are these tests commented out?

PR #105, opened 2018-03-13
Fast text embedder

dilyararimovna:
change fasttext(0.8.3) to fastText in requirements, fasttext embeddder class (pyfasttext is still here)
add error when n_classes is zero
change configs for intents
add new files "_v2" to share


deeppavlov/models/embedders/fasttext_embedder.py
```diff
@@ -50,14 +50,8 @@ def load(self, *args, **kwargs):
             log.info("[loading embeddings from `{}`]".format(self.load_path))
             model_file = str(self.load_path)
             if self.emb_module == 'fasttext':
-                import fasttext as Fasttext
+                import fastText as Fasttext
                 model = Fasttext.load_model(model_file)
```
yoptar: Can we throw this if out and do the import at the top of the file?

seliverstov: Done.

PR #117, opened 2018-03-16
Add TravisCI integration

nikolay-bushkov:


nikolay-bushkov:
35 minutes for build.... I think it is too long given our pull request frequency. One option is to wait for Kubernetes and Jenkins on our servers.

PR #123, opened 2018-03-20
Ranking1

puleon:


yoptar:
Would you please also rename the class and the registered name for the data iterator?

puleon:
rename the class and the registered name for the data iterator

PR #130, opened 2018-03-21
Odqa

my-master:
Major changes in *core*:

* Rename DatasetBasicIterator -> DatasetLearningIterator
* Add class DatasetFittingIterator
* Add fit_batches() method to Estimator
* Add fit_batches() to train.py
* Make DataReader optional in JSON config in train_model_from_config()
* Rename method batch_generator() -> gen_batch() in DatasetLearningIterator (former DatasetBasicIterator)
* Rename method iter_all() -> get_instances() in DatasetLearningIterator (former DatasetBasicIterator)

Changes in the core are made due to lack of functionality for fitting on batches, for parallel fitting on batches (should be solved later) and for non-availability of the iterator during the fitting process. Vocabs and iterators, as they are, are not useful for large datasets.
There are a few errors in the project architecture that should be addressed in the nearest future.

Other changes:
* Edit all project code to work with new core changes. Basically, these are minor changes.
* Add ODQA model classes (models.vectorizers.HashingTfidfVectorizer, models.odqa.TfidfRanker)

Tests passed except for the last test for Ranking model (not enough memory on my GPU).


deeppavlov/core/commands/utils.py
```diff
@@ -42,3 +42,13 @@ def get_deeppavlov_root() -> Path:
...
+
+
+def is_empty(d: Path) -> bool:
+    """
+    Check if directory is empty.
+    """
+    for x in d.iterdir():
+        if x.is_file():
+            return True
+    return False
```
yoptar: Directory is empty when there's a file?

deeppavlov/core/common/params.py
```diff
@@ -55,9 +55,8 @@ def from_params(params: Dict, **kwargs) -> Component:
             log.exception(e)
             raise e
 
-    try:
-        cls_name = config_params.pop('name')
-    except KeyError:
+    cls_name = config_params.pop('name', None)
+    if not cls_name:
```
yoptar: How is this better?

deeppavlov/core/data/data_learning_iterator.py
```diff
@@ -20,20 +20,24 @@
...
-        in each field.
+        """ Dataiterator takes a dict with fields 'train', 'test', 'valid'. A list of samples
+         (pairs x, y) is stored in each field.
         Args:
-            data: list of (x, y) pairs. Each pair is a sample from the dataset. x as well as y can be a tuple
-                of different input features.
-            seed (int): random seed for data shuffling. Defaults to None
+            data: list of (x, y) pairs. Each pair is a sample from the dataset. x as well as y
+            can be a tuple of different input features.
+            seed (int): random random for data shuffling. Defaults to None
```
yoptar: `random random` → `random seed`?

deeppavlov/core/data/data_learning_iterator.py
```diff
@@ -51,11 +55,10 @@ def __init__(self, data: Dict[str, List[Tuple[Any, Any]]],
             'all': self.train + self.test + self.valid
         }
 
-    def batch_generator(self, batch_size: int, data_type: str = 'train',
-                        shuffle: bool = None) -> Generator:
-        r"""This function returns a generator, which serves for generation of raw
-        (no preprocessing such as tokenization)
-         batches
+    def gen_batch(self, batch_size: int, data_type: str = 'train',
```
yoptar: maybe `generate_batches` then?

deeppavlov/core/commands/utils.py
my-master: Switch True and False :D

yoptar: And what if there's a directory, but no files?

deeppavlov/core/common/params.py
my-master: Shouldn't throw an exception, if there is a valid checking key method like "pop".

deeppavlov/core/data/data_learning_iterator.py
my-master: Yep.

deeppavlov/core/data/data_learning_iterator.py
my-master: Ok

deeppavlov/core/commands/utils.py
my-master: So you suggest what, recursive checking?

yoptar: You could return in first iteration without checking anything. or just `bool(list(d.iterdir()))`

deeppavlov/configs/odqa/ranker_prod.json
```diff
@@ -0,0 +1,21 @@
+{
+  "dataset_iterator": {
+    "name": "sqlite_iterator",
+    "data_path": "odqa",
+    "batch_size": 10000,
+    "data_url": "http://lnsigo.mipt.ru/export/datasets/wikipedia/wiki_full.db"
+  },
+  "model": {
```
seliverstov: All configs should be in the new format like  
"chainer": {
    "in": ["x"],
    "pipe": [ ...

seliverstov: It seems that due to old config format you model can't be deployed with REST API and your model can't be added to demo UI.

my-master: Current Chainer class is not applicable to my model. I can try to refactor the class, but then we won't deploy in time anyway. So what looks faster to you - refactoring Chainer or changing the deployment code so it would work with old configs?

seliverstov: Without unified config structure and with partially broken functions the model won't be added to release. It will stay in the current branch. I will make a patch to deployment code to run your model on demo stand.

my-master: Ok then, so I will refactor the Chainer class.

PR #144, opened 2018-03-26
Odqa Ranker

my-master:
1. The same as in pull-request #130 
2. Comments from pull-request #130 are taken into consideration.
3. New files in dev are changed according to the core refactoring, dev is merged.
3. Not all tests are passed (consecutive ner training and infering failed).

Update:
* fixed ner ontonotes consecutive training-inferring
* All tests passed.


deeppavlov/core/data/data_fitting_iterator.py
```diff
@@ -0,0 +1,68 @@
...
+        self.batch_size = batch_size
+        self.shuffle = shuffle
+        self.random = Random(seed)
+        self.data = data
+        self.doc_index = doc_ids or self.get_doc_ids()
+        self.data = data
+
+    def get_doc_ids(self):
+        ids = [i for i in range(len(self.data))]
+        return ids
```
yoptar: `list(range(len(self.data)))` is faster and better

deeppavlov/core/data/data_fitting_iterator.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def gen_batch(self, batch_size=1000, shuffle=False) -> Generator[Tuple[List[list],
+                                                                           List[int]], Any, None]:
+        _batch_size = self.batch_size or batch_size
```
yoptar: Argument to the method should have priority. One could set default value here to `None` and then
```
_batch_size = batch_size or self.batch_size or 1000
```
or just have default value for self.batch_size

deeppavlov/core/data/data_fitting_iterator.py
```diff
@@ -0,0 +1,68 @@
...
+        _shuffle = self.shuffle or shuffle
```
yoptar: Argument to the method should have priority. Also, since `shuffle` is boolean, it should be compared with proper `None` checks

deeppavlov/core/data/data_fitting_iterator.py
```diff
@@ -0,0 +1,68 @@
...
+
+        if _shuffle:
+            self.random.shuffle(self.doc_index)
```
yoptar: After getting shuffled data once, you loose the ability to get the data in original order.

deeppavlov/core/data/data_fitting_iterator.py
my-master: True, it is 10 ms faster per 1000000 iterations.

deeppavlov/core/data/data_fitting_iterator.py
my-master: Why would I need the original order?

yoptar: Why do you need shuffle flag if you don't need original order?

deeppavlov/core/data/data_fitting_iterator.py
my-master: Disagree.
When the instance is initialized, a user sets all needed arguments values (which are always None by default), and then these values are used in all methods. If they do not set the initial values, the values from the methods are taken. I could set default __init__ values to the methods' values, but I'd like to give a user an ability to use these methods standalone (though it is a very rare use case).
If anybody sets the `batch_size` or `shuffle` to None, then an error raises, and this is a proper behavior.

yoptar: When I specifically ask for `iterator.gen_batch(batch_size=3)` and get batches of 10, I'd think that something is wrong.

deeppavlov/core/data/data_fitting_iterator.py
yoptar: about a second faster for `range(100_000_000)`

deeppavlov/core/data/data_fitting_iterator.py
my-master: About priority - my prev comment.
About booleans. It would be an issue, when `self.shuffle=True`, but in this particular function a user wants `shuffle=False`. This would be impossible, because `_shuffle = True or False` will lead to `True` anyway. However, I don't see this would be a proper use case for this class. If users want `shuffle=False`, they should init a class with proper values. So I'd better raise a warning here, instead of allowing users mess with functionality as they please.

deeppavlov/core/data/data_fitting_iterator.py
my-master: If you want `batch_size = 3`, you should init the class with `batsh_size=3`.  And if you have a use case when you need to use `gen_batch()` always with different `batch_size`, then leave in init the default value, which is None.

deeppavlov/core/data/data_fitting_iterator.py
yoptar: Original order can be used to properly compare different ways of processing. You could train your model on shuffled data and then predict on ordered, for example.

yoptar: If you always want to shuffle, you do not need a flag for it, you can just always shuffle.

my-master: To compare different ways of processing you can shuffle with the same seed. 
" You could train your model on shuffled data and then predict on ordered, for example." - so set shuffle to True when training and set it to False when predicting?

yoptar: > so set shuffle to True when training and set it to False when predicting?   

Exactly. Preferably without the need to recreate the object.

my-master: For this use case just leave the __init__ value as `None` and set it to `True` or `False` when shuffling for training and predicting accordingly.
The logic is clear, there are 2 possible use cases:
1. You want to preserve any argument value in a class instance, init it with the value you need and don't worry about messing up with different values (it does not always perform well). 
2. You want to change arguments. Then leave the default value as ` None` and call methods with needed values.
Besides, with the current DeepPavlov's functionality you have to recreate the object anyway (consecutive training and interact modes recreate objects).

deeppavlov/core/data/data_fitting_iterator.py
seliverstov: I agree with @yoptar When `get batch(batch_size=3)` return batches of size other then 3 it's not appropriate behavior. Moreover, DataFitingIterator and DataLearningIterator should have consistent behavior: get batch(batch_size=3) should work exactly the same way on both iterators.

deeppavlov/core/data/data_fitting_iterator.py
seliverstov: Same point as in the previous comment.

deeppavlov/core/data/data_fitting_iterator.py
my-master:
```
# proper design:

class MyIterator:
    def __init__(self, data, batch_size=None):
        self.data = data
        self._batch_size = batch_size

    @property
    def batch_size(self):
        return self._batch_size

    @batch_size.setter
    def batch_size(self, batch_size):
        self._batch_size = batch_size

    def gen_batches(self):
        batches = [] # count with self._batch_size
        for i, doc_ids in enumerate(batches):
            pass # whatever


iterator = MyIterator(["data"], 23)
iterator.batch_size = 12
docs = iterator.gen_batches()


# inconvenient design:

class MyIterator:
    def __init__(self, data):
        self.data = data

    @property
    def batch_size(self):
        return self._batch_size

    @batch_size.setter
    def batch_size(self, batch_size):
        self._batch_size = batch_size

    def gen_batches(self, batch_size):
        batches = []  # count with batch_size
        for i, doc_ids in enumerate(batches):
            pass  # whatever


iterator = MyIterator(["data"])
docs = iterator.gen_batches(12)
```

my-master: `DataLearningIterator` is designed inconveniently, like the second one in the snippet above.  Instead of initializing the object once with proper parameters, and reassign them safely when needed, I have to mess up with bool and None checks inside the methods, and keep in mind what value of `batch_size` I have now and assign the same `batch_size` value in many places in my code to make sure that it is the same.

seliverstov: Agree. One way to eliminate inconsistent behavior is to remove `batch_size` parameter from `gen_batch` and use it only in `__init__` for both iterators: `data_learning_iterator` and `data_fitting_iterator`. I will accept such refactoring.

my-master: Ok then, I'll refactor both classes, run tests and update the pull-request.

my-master: I refactored only `DataFittingIterator` and didn't touch `DataLearningIterator` because it would affect a lot of functionality and people would become upset.
To refactor `DataLearningIterator` functionality we need not only to add properties as I did in `DataFittingIterator` but also:
1. Rewrite all inherited iterators to place `batch_size`, `shuffle`, etc. in their __init__ methods.
2. Rewrite configs to put `batch_size` not under `train_config`, but under `iterator` options.
3. Modify the `train.py` script, so there would be proper properties assigning and calling.

seliverstov:
`DataLearningIterator` still has 'batch_size` parameter for `get_batches` method and does not have the same parameter for `__init__` method. Both iterators should have identical behavior.

deeppavlov/core/commands/train.py
```diff
@@ -125,6 +144,8 @@ def train_model_from_config(config_path: str) -> None:
 
     if callable(getattr(model, 'train_on_batch', None)):
         _train_batches(model, iterator, train_config, metrics_functions)
+    elif callable(getattr(model, 'fit_batches', None)):
+        _fit_batches(model, iterator, train_config)
     elif callable(getattr(model, 'fit', None)):
         _fit(model, iterator, train_config)
```
yoptar: All estimators now have the `fit_batches` so this case block will never reach the `fit` part, but most of those estimators still want to fit the old way and just `pass` in the `fit_batches` method so training would not happen.

deeppavlov/core/data/data_fitting_iterator.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __init__(self, data: List[str], doc_ids: List[Any] = None,
+                 seed: int = None, shuffle: bool = True,
+                 *args, **kwargs) -> None:
+
+        self.shuffle = shuffle
+        self.random = Random(seed)
+        self.data = data
+        self.doc_index = doc_ids or self.get_doc_ids()
+        self.data = data
```
yoptar: `self.data = data` twice

deeppavlov/core/data/data_fitting_iterator.py
```diff
@@ -0,0 +1,67 @@
...
+        return list(range(len(self.data)))
+
+    def gen_batches(self, batch_size: int, shuffle: bool = None)\
+            -> Generator[Tuple[List[list], List[int]], Any, None]:
+
+        if shuffle is None:
+            shuffle = self.shuffle
+
+        if shuffle:
+            self.random.shuffle(self.data)
```
yoptar: Looks like it would be easier to shuffle index and not the data itself. Also, if you shuffle the data this way once, you loose it's original order. I would suggest something like
```
doc_index = list(self.doc_index)
self.random.shuffle(doc_index)
```

deeppavlov/dataset_iterators/sqlite_iterator.py
```diff
@@ -0,0 +1,108 @@
...
+        cursor.execute('SELECT id FROM {}'.format(self.db_name))
+        ids = [ids[0] for ids in cursor.fetchall()]
+        cursor.close()
+        return ids
+
+    def get_db_name(self) -> str:
+        cursor = self.connect.cursor()
+        cursor.execute("SELECT name FROM sqlite_master WHERE type='table';")
+        assert cursor.arraysize == 1
+        name = cursor.fetchmany(0)[0][0]
```
yoptar: So `name = cursor.fetchone()[0]`?

deeppavlov/dataset_iterators/sqlite_iterator.py
```diff
@@ -0,0 +1,108 @@
...
+        cursor.close()
+        return name
+
+    def map_doc2idx(self) -> Dict[int, Any]:
+        doc2idx = {doc_id: i for i, doc_id in enumerate(self.doc_ids)}
+        print("SQLite iterator: The size of the database is {} documents".format(len(doc2idx)))
```
yoptar: Please, use logging

deeppavlov/dataset_iterators/sqlite_iterator.py
```diff
@@ -0,0 +1,108 @@
...
+            (doc_id,)
+        )
+        result = cursor.fetchone()
+        cursor.close()
+        return result if result is None else result[0]
+
+    @overrides
+    def gen_batches(self, batch_size=1000, shuffle=False) -> Generator[Tuple[List[str], list], Any, None]:
+        _batch_size = self.batch_size or batch_size
+        _shuffle = self.shuffle or shuffle
```
yoptar: If `shuffle` is explicitly `False`, but `self.shuffle` is `True`, this will be `True`

deeppavlov/dataset_iterators/sqlite_iterator.py
```diff
@@ -0,0 +1,108 @@
...
+
+        if _shuffle:
+            self.random.shuffle(self.doc_ids)
```
yoptar: Original order is lost

deeppavlov/dataset_iterators/sqlite_iterator.py
```diff
@@ -0,0 +1,108 @@
...
+        print("SQLite iterator: The size of the database is {} documents".format(len(doc2idx)))
+        return doc2idx
+
+    def get_doc_content(self, doc_id: Any) -> Optional[str]:
+        cursor = self.connect.cursor()
+        cursor.execute(
+            "SELECT text FROM documents WHERE id = ?",
+            (doc_id,)
+        )
+        result = cursor.fetchone()
```
yoptar: It should be faster to select in batches and then just sort the result in desired order

deeppavlov/dataset_iterators/sqlite_iterator.py
my-master: Refactored

deeppavlov/core/data/data_fitting_iterator.py
my-master: Refactored

deeppavlov/dataset_iterators/sqlite_iterator.py
my-master: Refactored

deeppavlov/models/vectorizers/hashing_tfidf_vectorizer.py
```diff
@@ -0,0 +1,173 @@
...
+        self.doc_index = doc_index
+
+        global TOKENIZER
+        TOKENIZER = self.tokenizer
+
+        self.rows = []
+        self.cols = []
+        self.data = []
+
+    def __call__(self, question: str) -> sp.sparse.csr_matrix:
```
yoptar: Components are expected to work with batches, not individual instances

deeppavlov/configs/odqa/ranker_prod.json
```diff
@@ -0,0 +1,52 @@
+{
+  "dataset_iterator": {
+    "name": "sqlite_iterator",
+    "data_dir": "odqa",
+    "shuffle": false,
+    "data_url": "http://lnsigo.mipt.ru/export/datasets/wikipedia/wiki_full.db"
```
yoptar: This file is 13GB but it can be gzipped to 4.3GB

my-master: Yes, but this will take too much time to decompress. So the question is, what we want to save for our users, time or traffic?

yoptar: I decompressed it in a minute and a half. Usually it takes more than that to download 8GB.

seliverstov: Let's try to save traffic.

deeppavlov/models/vectorizers/hashing_tfidf_vectorizer.py
my-master: Done

deeppavlov/configs/odqa/ranker_prod.json
my-master: On my machine it takes much longer to decompress (about 7 mins)
```
2018-04-06 12:14:08.779 INFO in 'deeppavlov.core.data.utils'['utils'] at line 54: Downloading from http://lnsigo.mipt.ru/export/deeppavlov_data/odqa.tar.gz to /media/olga/Data/projects/iPavlov/DeepPavlov/download/odqa.tar.gz
100%|██████████████████████████████████████████████████████████████████████████████████| 4.53G/4.53G [01:23<00:00, 54.0MB/s]
2018-04-06 12:15:32.681 INFO in 'deeppavlov.core.data.utils'['utils'] at line 120: Decompressing...
2018-04-06 12:21:49.946 INFO in 'deeppavlov.core.data.utils'['utils'] at line 118: Extracting /media/olga/Data/projects/iPavlov/DeepPavlov/download/seq2seq_go_bot.tar.gz archive into /media/olga/Data/projects/iPavlov/DeepPavlov/download

```

my-master: Compressed.

deeppavlov/dataset_iterators/sqlite_iterator.py
my-master: For now, I postpone this task until merge. It doesn't work as easy as I expected, sqlite crashes and I need to investigate why.

PR #156, opened 2018-03-28
feat: Ontonotes NER added

mu-arkhipov:


deeppavlov/models/ner/network_ontonotes.py
```diff
@@ -0,0 +1,187 @@
...
+
+        # Mask for paddings
+        x['mask'] = np.zeros([batch_size, max_utt_len], dtype=np.float32)
+        for n in range(batch_size):
+            x['mask'][n, :len(batch_x[n])] = 1
+
+        return x
+
+    def train_on_batch(self, x_word, x_char, y_tag):
+        raise NotImplementedError
```
yoptar: If it cannot be trained it could just be a serializable and not an NNModel. And a config for it could not have an `in_y` field at all

PR #175, opened 2018-04-06
Dialog

aCombray:
Implement a rule based slot filler. It uses Levenshtein distance to extract the potential entities. We also provide two configurations that use this filler: gobot_my.json and gobot_simple.json. Both require wiki fasttext downloaded from [fasttext wiki](https://github.com/facebookresearch/fastText/blob/master/pretrained-vectors.md)
Use configuration gobot_my.json, the accuracy is 
```
{"valid": {"examples_seen": 575, "metrics": {"per_item_dialog_accuracy": 0.530427954456223}, "time_spent": "0:02:14"}}
{"test": {"examples_seen": 576, "metrics": {"per_item_dialog_accuracy": 0.5310756205503175}, "time_spent": "0:02:00"}}
```
Use configuration gobot_simple.json, the accuracy is 
```
{"valid": {"examples_seen": 575, "metrics": {"per_item_dialog_accuracy": 0.5231645072634472}, "time_spent": "0:01:54"}}
{"test": {"examples_seen": 576, "metrics": {"per_item_dialog_accuracy": 0.5320377140658072}, "time_spent": "0:01:52"}}
```


aCombray:
Both tests are on the dstc2 data. But I don't know whether it is the turn accuracy or action accuracy.

vikmary:
@aCombray "per_item_dialog_accuracy" metric in gobot configs correspond to turn accuracy (for each turn lower cased predicted and true texts are compared).

mu-arkhipov:
Could you please rename the modules contaning my_component or something like this to pharma_bot, rule_based_slotfill and so on? It would provide more information to someone who see it first time

aCombray:
I added modules for retrieving data from a sql file. It is the bot_sql in skills/go_bot, new template class SQLTemplate, a class for retrieving the data, sql_database_reader

mu-arkhipov:
Could you share the data you use?

aCombray:
Unfortunately, I can not share the data because of copyright issue. I can provide an example here. The template is a list of three words like the following: action_label,  cur.execute("SQL command"), answer_text = cur.fetchall()[0][0]. The second item is the execution of the SQL command, and the third item is to get the result. I think in practice, it is better to allow users to write their own data reader class with their own database.

mu-arkhipov:
Thank you for your contribution. In days to come we will merge this branch into our development branch.

vikmary:
@aCombray We wouldn't be able to keep configs related to pharma, because dataset is unavailable. But

1. simple_slot_filler
2. and sql database component (with a bunch of refactors)

will be contributed. Thanks for your work!

The following remark might be helpful: your [`deeppavlov/configs/go_bot/gobot_pharma2.json`](https://github.com/aCombray/DeepPavlov/blob/fb60cadba3555c01ba0dda25be3a7e2a39e0c022/deeppavlov/configs/go_bot/gobot_pharma2.json) uses intents trained on dstc2-dataset (as it seems to me). If my guess was right, you might improve your scores by training your own pharma intent classifier (for details see [`deeppavlov/models/classifiers/intents`](https://github.com/aCombray/DeepPavlov/tree/master/deeppavlov/models/classifiers/intents)).

PR #190, opened 2018-04-16
ODQA model

my-master:
This pull-request includes the following main changes:
* ODQA test config (infer only)
* ODQA prod config (infer only)
* odqa moved to skills
* wiki_sqlite vocab (a connection component between ranker and reader)
* ODQA test config included in tests

**Update**
* remove start answer position from the final ODQA output in prod config
* enable batch processing in WikiSqlite vocab


deeppavlov/vocabs/wiki_sqlite.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+
+    def __init__(self, data_dir: str = '', data_url: str = DB_URL, **kwargs):
+        """
+        :param data_dir: a directory name where DB is located
+        :param data_url: an URL to SQLite DB
+        """
+        download_dir = expand_path(data_dir)
+        download_path = download_dir.joinpath(data_url.split("/")[-1])
+        download(download_path, data_url, file_exists=False)
```
yoptar: Can we rename `file_exists` to something like `force_download` or `redownload`?

deeppavlov/core/data/utils.py
```diff
@@ -91,7 +91,7 @@ def ungzip(file_path, extract_folder=None):
             extract_folder: folder to which the files will be extracted
 
         """
-    CHUNK = 16 * 1024
+    CHUNK = 2 ** 14
```
yoptar: I think that 16 * 1024 is more readable

my-master: Done

deeppavlov/vocabs/wiki_sqlite.py
my-master: Done

PR #192, opened 2018-04-17
SQuAD model for Russian

yurakuratov:
- model for Russian
- learning_rate decay
- weight decay


deeppavlov/core/layers/tf_layers.py
```diff
@@ -241,10 +316,10 @@ def stacked_highway_cnn(units: tf.Tensor,
                                  filter_width,
                                  padding='same',
                                  dilation_rate=dilation_rate,
-                                 kernel_initializer=xavier_initializer())
+                                 kernel_initializer=INITIALIZER)
         if use_batch_norm:
             units = tf.layers.batch_normalization(units, training=training_ph)
-        sigmoid_gate = tf.layers.dense(input_units, 1, activation=tf.sigmoid, kernel_initializer=xavier_initializer())
+        sigmoid_gate = tf.layers.dense(input_units, 1, activation=tf.sigmoid, kernel_initializer=INITIALIZER)
```
yoptar: `INITIALIZER` is called in some places but not in others

PR #196, opened 2018-04-23
Spacy tokenizer

my-master:
Non-streaming SpaCy tokenizer. Add it because some people complain about slow working of streaming tokenizer. On some machines SpaCy threads behave oddly.
* non streaming spacy tokenizer
* rename spacy_tokenizer.py -> streaming_spacy_tokenizer.py


yoptar:
Those two have a lot in common. Maybe inherit one from the other or even leave one and parametrize it somehow?

my-master:
Definitly should do a superclass for all tokenizers. At least __call__() and _filter() methods are identical. Also some __init__ attributes. Again, can do this only after May 7.

PR #199, opened 2018-04-24
Feature/gobot with new intents

vikmary:
- [x] nltk_tokenizer returns list of tokens, not str
- [x] intent_classifier and error_model expect list of tokens from nltk_tokenizer
- [x] retrain model with new intents, upload on share and modify `download.py`


deeppavlov/configs/intents/intents_dstc2.json
```diff
@@ -22,65 +22,86 @@
...
+        "model_name": "cnn_model",
+        "embedder": "#my_embedder",
+        "tokenizer": "#my_tokenizer"
       }
     ],
-    "out": ["y_predicted"]
+    "out": [
+      "y_labels",
+      "y_probas",
+      "y_classes"
```
yoptar: It's weird that classes are returned as a model's output in every batch. Will it be bad if probabilities will be returned in the form of `dict(zip(y_classes, y_probas))`?

dilyararimovna: done

PR #220, opened 2018-05-07
Docs/intents

dilyararimovna:
add info to README about models zoo


seliverstov:
Also, fix list of new models for `model_name` at `parameters for building the main part of a model`

PR #224, opened 2018-05-10
Feature/two intent models dstc2

dilyararimovna:
add another config for dstc2 with bigger embeddings, fixed readme


vikmary:
Maybe we should also mention how much operative memory is required?

PR #229, opened 2018-05-14
Feature / single one ner

mu-arkhipov:


deeppavlov/dataset_readers/ontonotes_reader.py
```diff
@@ -0,0 +1,49 @@
...
+
+@register('ontonotes_reader')
+class OntonotesReader(DatasetReader):
+    URL = 'http://lnsigo.mipt.ru/export/datasets/ontonotes_senna.pckl'
+
+    def read(self, data_path, file_name: str='ontonotes_senna.pckl', provide_senna_pos=False, provide_senna_ner=False):
+        path = Path(data_path).resolve() / file_name
+        if not path.exists():
+            download_decompress(self.URL, str(path.parent))
+        with open(path, 'rb') as f:
```
yoptar: Will break here if file_name is not default

deeppavlov/models/ner/network.py
```diff
@@ -267,6 +270,24 @@ def __call__(self, *args, **kwargs):
...
+                self._best_f1 = 0
+            if not hasattr(self, '_impatience'):
+                self._impatience = 0
+            if data['metrics']['ner_f1'] > self._best_f1:
+                self._best_f1 = data['metrics']['ner_f1']
+                self._impatience = 0
+            else:
+                self._impatience += 1
+
+            if self._impatience > 4:
```
yoptar: lr_patience could be a parameter

PR #243, opened 2018-05-21
Ranking new interfaces

puleon:


deeppavlov/core/data/utils.py
```diff
@@ -23,7 +23,6 @@
 import numpy as np
 import re
 import zipfile
-import shutil
```
yoptar: But why? `shutil.copy` is used in this file

puleon: fixed.

PR #249, opened 2018-05-25
Fix/intents params

dilyararimovna:
add re-init of changable params for keras model, insults as sentiment


deeppavlov/configs/sentiment/insults_kaggle.json
```diff
@@ -0,0 +1,135 @@
...
+        "subdir": "insults_data"
+      },
+      {
+        "url": "http://lnsigo.mipt.ru/export/datasets/insults_data/valid.csv",
+        "subdir": "insults_data"
+      },
+      {
+        "url": "http://lnsigo.mipt.ru/export/datasets/insults_data/test.csv",
+        "subdir": "insults_data"
+      },
```
yoptar: You can add `train.csv`, `valid.csv` and `test.csv` into a single `tar.gz` archive

PR #253, opened 2018-05-31
Cove embedder

fogside:
Added bpe tokenizer + cove_embedder


deeppavlov/models/embedders/cove_embedder.py
```diff
@@ -0,0 +1,150 @@
...
+        if self.reduce_method == 'mean':
+            encoded = np.mean(outputs, axis=1)
+        elif self.reduce_method == 'sum':
+            encoded = np.sum(outputs, axis=1)
+        elif self.reduce_method == 'max':
+            encoded = np.max(outputs, axis=1)
+        elif self.reduce_method == 'none':
+            encoded = outputs
+        else:
+            log.error("None of reducing methods has been defined")
```
yoptar: Can this be resolved in `__init__`? You can make `self.reduce_method` a method there and then just call
```
encoded = self.reduce_method(outputs)
```
or something like that

deeppavlov/models/embedders/cove_embedder.py
```diff
@@ -0,0 +1,150 @@
...
+from deeppavlov.core.models.tf_backend import TfModelMeta
+from typing import List
+
+log = get_logger(__name__)
+
+
+@register('cove_embedder')
+class CoVeEmbedder(Component, metaclass=TfModelMeta):
+    def __init__(self, vocab_file_path: str, checkpoint_path: str, num_layers: int, cell_class_name: str,
+                 num_units: int, residual_connections: bool, reduce_method: str, **kwargs):
```
yoptar: I feel that some of these parameters can have default values.

deeppavlov/models/embedders/cove_embedder.py
fogside: Yes, agree with it

deeppavlov/models/tokenizers/bpe_tokenizer.py
```diff
@@ -0,0 +1,180 @@
...
+                self._decode = self.model.DecodePieces
+        else:
+            log.error("Trying to use self.model which doesn't exist. Please, fit model before using.")
+            exit(1)
+
+    def __len__(self):
+        return self.vocab_size
+
+    def save(self, *args, **kwargs):
+        raise NotImplementedError
```
yoptar: save is always called after fit

PR #285, opened 2018-06-25
Feature/params evolution

dilyararimovna:
parameters evolution


deeppavlov/evolve.py
```diff
@@ -0,0 +1,322 @@
...
+parser.add_argument('--p_size', help='population size', type=int, default=10)
+parser.add_argument('--gpus', help='visible GPUs divided by comma <<,>>', default="-1")
+parser.add_argument('--train_partition',
+                    help='partition of splitted train file', default=1)
+parser.add_argument('--start_from_population',
+                    help='population number to start from. 0 means from scratch', default=0)
+parser.add_argument('--path_to_population',
+                    help='path to population to start from', default="")
+parser.add_argument('--elitism_with_weights',
+                    help='whether to save elite models with weights or without', default=0)
```
yoptar: maybe use `action='store_true'` for `elitism_with_weights`?

deeppavlov/evolve.py
```diff
@@ -0,0 +1,322 @@
...
+
+    if os.environ.get("CUDA_VISIBLE_DEVICES") is None:
+        pass
+    else:
+        cvd = [int(gpu) for gpu in os.environ.get("CUDA_VISIBLE_DEVICES").split(",")]
+        if set(gpus).issubset(set(cvd)):
+            pass
+        else:
+            try:
+                gpus = [cvd[gpu] for gpu in gpus]
```
yoptar: I think that `gpus = [cvd[gpu] for gpu in gpus]` should be used even `if set(gpus).issubset(set(cvd))`

deeppavlov/core/commands/train.py
```diff
@@ -188,6 +190,9 @@ def train_evaluate_model_from_config(config: [str, Path, dict], to_train=True, t
             }
 
             print(json.dumps(report, ensure_ascii=False))
+            all_reports.append(report)
+
+    return all_reports
```
yoptar: This will return a `NameError` `if not (train_config['validate_best'] or train_config['test_best'])`

deeppavlov/evolve.py
```diff
@@ -0,0 +1,322 @@
...
+    population_scores = results_to_table(population, evolution, considered_metrics,
+                                         result_file, result_table_columns)[evolve_metric]
+    log.info("Population scores: {}".format(population_scores))
+    log.info("Iteration #{} was done".format(iters))
+    iters += 1
+
+    while True:
+        if iters == iterations:
+            log.info("End of evolution on iteration #{}".format(iters))
+            break
```
yoptar: What if you set iteration limit to 5 and start from 7th?

deeppavlov/evolve.py
dilyararimovna: I don't think so, because all of the files are actually stored. this parameter is responsible for translating elite models with or without weights from population to popultation

deeppavlov/evolve.py
dilyararimovna: fixed

deeppavlov/core/commands/train.py
dilyararimovna: fixed

deeppavlov/evolve.py
dilyararimovna: ok

deeppavlov/evolve.py
dilyararimovna: if set(gpus).issubset(set(cvd)), user means use 3,5 out of 3,5,7 visible devices.
else user means use 0,1 out of 3,5,6 visible devices. 
Used devices will be the same but the way of assigning is different.

dilyararimovna: for the second comment, it will raise error for this case because it is really not clear what the user wants

dilyararimovna: the second is fixed. it will raise error because it is unclear what to use

PR #304, opened 2018-07-06
Tfidf ranker moved to ranking

my-master:
* moved `tfidf_ranker` to `ranking` module
* included ODQA and ranker to tests (were excluded)
* moved ODQA and ranker test configs to tests
* created a separate README_TFIDF.md in `ranking` module
* edited ODQA README.md


my-master:
At this point there are some issues that should be addressed:
1. Not clear where a skill readme file should be if our skill exists only at json config level and not as a separate project module.
2. There is a mess in ranking module now and will be even bigger mess when i'll put their my other rankers and utils. Can be solved by creating separate folders for insurance ranker/utils. and for odqa rankers/utils.

Issue #301, opened 2018-07-04
Add an option to cache downloads


yoptar:
Tests are taking too long and are sometimes randomly failing on incomplete downloads. Adding a `.cache` directory where filenames are hashes of urls for example could resolve this.

PR #307, opened 2018-07-09
feat: add cache option for downloads to try and speed the tests up

yoptar:
- [x] cache downloads
- [x] cache extracted


yoptar:
resolves #301

PR #327, opened 2018-07-16
Tagger dev merge

AlexeySorokin:
Implements a better morphological tagging model which uses Pymorphy as an additional input to the model,


deeppavlov/core/common/chainer.py
```diff
@@ -143,6 +143,7 @@ def _predict(self, x, y=None, to_return=None):
         res = [mem[k] for k in to_return]
         if len(res) == 1:
             return res[0]
+        ### зачем??????
```
yoptar: remove this please

deeppavlov/models/morpho_tagger/tagger.py
```diff
@@ -104,15 +104,22 @@ def train_on_batch(self, x: list, y: list):
...
-
-    def __call__(self, x_batch, **kwargs):
+        if len(args) > 2:
+            # data, labels = list(args[:-1]), args[-1]
+            data, labels = [list(x) for x in args[:-1]], list(args[-1])
+        else:
+            data, labels = args
+        self._net.train_on_batch(data, labels, **self.train_parameters)
+
+    def __call__(self, x_batch, *args, **kwargs):
```
yoptar: maybe something like `def __call__(self, *x_batch, **kwargs):`?

deeppavlov/models/morpho_tagger/tagger.py
```diff
@@ -104,15 +104,22 @@ def train_on_batch(self, x: list, y: list):
         Returns:
 
         """
-        self._net.train_on_batch(x, y, **self.train_parameters)
-
-    def __call__(self, x_batch, **kwargs):
+        if len(args) > 2:
+            # data, labels = list(args[:-1]), args[-1]
```
yoptar: This line can be safely removed, i think

deeppavlov/models/vectorizers/dictionary_vectorizer.py
```diff
@@ -0,0 +1,253 @@
...
+        self.unk_token = unk_token
+        self.load()
+
+    @property
+    def dim(self):
+        return len(self._t2i)
+
+    def save(self):
+        save_path = str(self.save_path)
+        with open(save_path, "w", encoding="utf8") as fout:
```
yoptar: you can do `save_path.open('w', ...)` without casting to `str`

deeppavlov/models/vectorizers/dictionary_vectorizer.py
```diff
@@ -0,0 +1,253 @@
...
+            for word, curr_labels in sorted(self.word_tag_mapping.items()):
+                curr_labels = [self._i2t[index] for index in curr_labels]
+                curr_labels = [x for x in curr_labels if x != self.unk_token]
+                fout.write("{}\t{}".format(word, " ".join(curr_labels)))
+        return self
+
+    def load(self):
+        if isinstance(self.load_path, str):
```
yoptar: It can (and should) also be a `Path`, not only `str`

deeppavlov/models/vectorizers/dictionary_vectorizer.py
```diff
@@ -0,0 +1,253 @@
...
+            load_path = pathlib.Path(self.load_path)
+            if load_path.is_dir():
+                load_path = [str(x) for x in load_path.iterdir() if x.is_file()]
+            else:
+                load_path = [str(load_path)]
+        else:
+            load_path = [str(x) for x in self.load_path]
```
yoptar: no need to cast it to `str`

deeppavlov/models/vectorizers/dictionary_vectorizer.py
```diff
@@ -0,0 +1,253 @@
...
+                for line in fin:
+                    line = line.strip()
+                    if line.count("\t") != 1:
+                        continue
+                    word, labels = line.split("\t")
+                    labels_by_words[word].update(labels.split())
+        self._train(labels_by_words)
+        return self
+
+    def _train(self, labels_by_words : Dict):
```
yoptar: If it's trainable, should it not be an Estimator?

deeppavlov/models/vectorizers/dictionary_vectorizer.py
```diff
@@ -0,0 +1,253 @@
...
+        max_length = max(len(x) for x in data)
+        answer = np.zeros(shape=(len(data), max_length, self.dim), dtype=int)
+        for i, sent in enumerate(data):
+            for j, word in enumerate(sent):
+                answer[i, j][self.word_tag_mapping[word]] = 1
+        return answer
+
+
+@register("pymorphy_vectorizer")
+class PymorphyVectorizer(Serializable):
```
yoptar: It shares `__call__` method with `DictionaryVectorizer`, maybe inherit one from another?

dp_requirements/morpho_tagger.txt
```diff
@@ -0,0 +1 @@
+russian-tagsets
```
yoptar: It's better to use a specific version, as changes can break compatibility (or a package itself)

deeppavlov/models/vectorizers/dictionary_vectorizer.py
AlexeySorokin: Renamed to _initialize since _train does simple loading.

deeppavlov/models/vectorizers/dictionary_vectorizer.py
AlexeySorokin: They call a different function inside cycle, however, it seems reasonable to create a basic class for both vectorizers.

PR #330, opened 2018-07-18
feat: simplify spacy en model installation

yoptar:


yoptar:
https://spacy.io/usage/models#usage-import
How you choose to load your models ultimately depends on personal preference. However, for larger code bases, we usually recommend native imports, as this will make it easier to integrate models with your existing build process, continuous integration workflow and testing framework. It'll also prevent you from ever trying to load a model that is not installed, as your code will raise an ImportError immediately, instead of failing somewhere down the line when calling spacy.load().

We need to hardcode it for the same reason we hardcode versions in all our requiremets: to ensure that everything works and works the same way as when we tested it.  
Also, the subprocess version is a hack, and we already have a hack for tensorflow that we cannot get rid of.

my-master:
Still, it makes impossible to use other than en_core_web_sm models. It makes some sense for production but doesn't for research. What if a researcher needs to reproduce exact article experiments and another model was used there? They will have to change the source code to be able to do that.

my-master:
Unfortunately, the current dev version with that subprocess thing - yes, and I didn't vote for that solution.

vikmary:
Verified that `go_bot` model works fine with new spacy installation.
`seq2seq_bot` also uses spacy, does it also need en_core_web_sm requirements' file?

PR #369, opened 2018-08-10
Log/prints to logs

yoptar:


deeppavlov/models/classifiers/keras_classification_model.py
```diff
@@ -133,7 +133,10 @@ def __init__(self, text_size: int,
             if self.opt['fasttext_md5'] != current_fasttext_md5:
                 raise ConfigError(
                     "Given fasttext model does NOT match fasttext model used previously to train loaded model")
-        print("Model was successfully initialized!\nModel summary:\n{}".format(self.model.summary()))
+
+        summary = ['Model was successfully initialized!', 'Model summary:']
+        self.model.summary(print_fn=summary.append)
```
vikmary: Can't we just do the logging in a single line?
`log.info("Model was successfully initialized!\nModel summary:\n{}".format(self.model.summary()))`

yoptar: `def summary(print_fn=print)` returns nothing and executes `print_fn` on every line of the summary, so it will not work like that

PR #377, opened 2018-08-16
Fix/params evolution refactor

dilyararimovna:
extract ParamsSearch from ParamsEvolution for further use in ParamsGridSearch, ParamsRandomSearch


deeppavlov/core/common/params_search.py
```diff
@@ -0,0 +1,250 @@
...
+from typing import List, Generator, Tuple, Any
+
+from deeppavlov.core.common.registry import register
+from deeppavlov.core.common.log import get_logger
+
+
+log = get_logger(__name__)
+
+
+@register('params_search')
```
yoptar: No need to register classes that will not be referenced in config files

PR #382, opened 2018-08-17
Feature/driver dataset gobot

vikmary:
- [x] implemented almost all features of [key-value networks](https://arxiv.org/pdf/1705.05414.pdf) (attention over knowledge base, encoder attention, embeddings instead of one-hots)
- [x] released a seq2seq_gobot_v2.tar.gz model with `BLEU=13.2`
- [x] updated doc


deeppavlov/core/models/tf_model.py
```diff
@@ -101,11 +101,17 @@ def get_train_op(self,
...
+                    if grad is None:
+                        return grad
+                    return tf.clip_by_norm(grad, clip_norm)
+
                 opt = optimizer(learning_rate)
                 grads_and_vars = opt.compute_gradients(loss, var_list=variables_to_train)
                 if clip_norm is not None:
-                    grads_and_vars = [(tf.clip_by_norm(grad, clip_norm), var)
-                                      for grad, var in grads_and_vars] #  if grad is not None
+                    grads_and_vars = [(clip_if_not_none(grad), var)
```
mu-arkhipov: This is a little bit tricky part. Sometimes error produced by the `tf.clip_by_norm` when feeding `None` is a sign of bugs in code. However, I would like to skip `None` grads.

deeppavlov/core/models/tf_model.py
```diff
@@ -101,11 +101,17 @@ def get_train_op(self,
             # For batch norm it is necessary to update running averages
             extra_update_ops = tf.get_collection(tf.GraphKeys.UPDATE_OPS)
             with tf.control_dependencies(extra_update_ops):
+                
+                def clip_if_not_none(grad):
```
mu-arkhipov:
```
def clip_if_not_none(grad):
   if grad is not None:
      return tf.clip_by_norm(grad, clip_norm)
```

deeppavlov/core/data/simple_vocab.py
```diff
@@ -48,9 +50,11 @@ def __init__(self,
         if self.load_path:
             self.load()
 
-    def fit(self, tokens):
+    def fit(self, *args):
```
mu-arkhipov: Why args?

deeppavlov/models/seq2seq_go_bot/kb_attn_layer.py
```diff
@@ -0,0 +1,180 @@
...
+from tensorflow.python.layers import base
+from tensorflow.python.ops import init_ops
+from tensorflow.python.framework import tensor_shape
+
+
+class KBAttention(base.Layer):
+# TODO: update class doc
+    """Densely-connected layer class.
+    Arguments:
+        units: Integer or Long, dimensionality of the output space.
```
mu-arkhipov: `kb_inputs`, `kb_mask`?

deeppavlov/core/data/simple_vocab.py
vikmary: Right. To be able to pass several variables to `fit()`.

PR #387, opened 2018-08-21
Requirements version updates + added spacy

arogozin:
This version removes installation problem `await = None ^ SyntaxError: invalid syntax`


seliverstov:
Hi,
Thanks. But we have separate requirements file for `spacy` and do not install it by default. Please remove spacy from requirements.txt

PR #415, opened 2018-08-30
Docs/core

nikolay-bushkov:


my-master:
Typing is completely incompetent and, therefore, not informative in this pull-request:
- Mixed "dict" and "Dict", "list" and "List" notations.
- Somewhere "Union" is not applied to enumeration, and somewhere is applied. Actually, should be always applied, if it is an enumeration of possible types.
- Somewhere "Optional" is accompanied with redundant "Union".
- Wrong handling (not like in pep8 recommendations) of mutable types in signatures.

See pep 8, pep 484 for detailed information.

nikolay-bushkov:
@my-master I see significant improvements in this pull request over the previous state of the repository.
I haven't found an appropriate explanation of what I was doing wrong, so either elaborate on this (provide some links more specific than the number of PEP), or make your own commits.
I am inclined to make iterative changes, and consider this iteration completed given such constraints as release schedule.

I have made changes in signatures and typing mainly in places near those where I was editing/adding docstrings.
I use built-in types where possible and make imports from typing (of List or Dict) only when there is some need to be more specific with types of elements.

my-master:
Almost everything I pointed out, can be easily found in the two mentioned peps, 8 and 484. Both of them (if you take only typing from pep 8) take about an hour to read, so I don't understand how I can be more specific here. All of your typing misunderstandings are literally discussed in these two peps.

1. `List` or `list`, `Dict` or `dict`. Since python 3.5 should always be not a type, but a _type name_

**pep 484:**

> When Type is parameterized it requires exactly one parameter. Plain Type without brackets is equivalent to Type[Any] and this in turn is equivalent to type (the root of Python's metaclass hierarchy). This equivalence also motivates the name, Type, as opposed to alternatives like Class or SubType, which were proposed while this feature was under discussion; this is similar to the relationship between e.g. List and list.

2. Enumeration just in brackets, without pointing a specific type (not sure this is your code, however, you edited these docstrings). You can't just simply write `[str, Path dict]`, this construction is meaningless. What is it, an enumeration like `Union`, a `List` that can contain specified elements, a `Tuple` with exact specified elements?

3. Redundant `Union` altogether with `Optional` in `Optional[Union[str, Path]] = None`

**pep 484**

> Optional, defined by Optional[t] == Union[t, type(None)]

Also, look the whole [section](https://www.python.org/dev/peps/pep-0484/#id29) about these types.

4. Mutable types in function signatures. You can find multiple explanations on StackOverflow why it shouldn't be done, so if you need to pass a list with default values to a signature, you should do it like it is explained in:

See nice explanation at [The Hitchhacker's Guide to Python](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). Also at [Python Anti-Patterns](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html).

Now, let's discuss how you've handled a mutable type in a signature. You converted it to `Tuple`:
`Optional[Iterable] = ('Optimizer',)`. `Tuple`, as an immutable, "freezed" data type is not only wrong for this variable semantics (it should be `List` or `Set`), but also `Iterable` shouldn't be used here at all. Someone can pass a `Dict` here, which leads to an error. 

5. See p. 4 about `Iterable`.

PR #446, opened 2018-09-13
Fix / Multiple keras models in parallel

vikmary:
Encapsulation of all KerasModel methods in graph and session for KERAS_BACKEND=tensorflow.

### DIFF
--- a/deeppavlov/core/models/tf_backend.py
+++ b/deeppavlov/core/models/tf_backend.py
@@ -20,6 +20,7 @@ import tensorflow as tf
 
 
 def _graph_wrap(func, graph):
+    """Constructs function encapsulated in the graph."""
     @wraps(func)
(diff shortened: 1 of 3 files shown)


deeppavlov/core/models/tf_backend.py
```diff
@@ -20,28 +20,58 @@
...
+                return obj
+
             K.clear_session()
+            obj.graph = tf.Graph()
+            with obj.graph.as_default():
+                if hasattr(cls, '_config_session'):
+                    obj.sess = cls._config_session()
+                else:
+                    obj.sess = tf.Session()
+                K.set_session(obj.sess)
```
mu-arkhipov: What is the reason to put another `K.set_session` here?
We already have it in every method.

vikmary: No reason really.

PR #447, opened 2018-09-13
chore: add possibility of computing accuracy for multidimensional predictions

PeganovAnton:
Consider seq2seq generating several sentences. y_predicted is list of lists of integers or
2D numpy.ndarray. When language model task is trained y_predicted, has shape
 [sequence_length, batch_size]. Old  function does not allow by-token accuracy calculation
in these cases.


yoptar:
`accuracy` function is used for simplest object to object comparisons. There are other variants of accuracy available in the same file, including `per_token_accuracy`.

PR #451, opened 2018-09-14
New python API for NER + Doc fixes

mu-arkhipov:
New, neat python API for NER models, check this out:
```python
from deeppavlov.models import ner
ner_model = ner.train_model('ner_ontonotes', 'dataset/path/') # path is optional
ner_model(['Your example to tag'])
```

or something like this for pre-trained models:
```python
from deeppavlov.models import ner
ner_model = ner.load_model('ner_ontonotes')
ner_model(['Sample to tag.'])
```
and for evaluation:
```python
from deeppavlov.models import ner
ner.evaluate_model('ner_ontonotes', 'dataset/path/') # path is optional
```


deeppavlov/core/commands/train.py
```diff
@@ -194,6 +194,7 @@ def train_evaluate_model_from_config(config: [str, Path, dict], to_train: bool =
             }
 
             print(json.dumps(report, ensure_ascii=False))
+    return model
```
yoptar: If an early stopping occurred but the model was not tested, this will return not a best model.

deeppavlov/download.py
```diff
@@ -37,10 +39,15 @@
...
     dp_root_back = get_deeppavlov_root()
-    config = read_json(config_path)
-    set_deeppavlov_root(config)
+    if isinstance(config, (str, Path)):
+        config = read_json(config)
+        set_deeppavlov_root(config)
+    else:
+        deeppavlov_root = Path('download/')
+        deeppavlov_root.mkdir(exist_ok=True)
+        paths.deeppavlov_root = deeppavlov_root
```
yoptar: This is very questionable. The config could have the `deeppavlov_root` property set but it is ignored here.

seliverstov:
Implemented at PR #551

PR #458, opened 2018-09-19
feat: add possibility of validation after N batches

PeganovAnton:
Tasks with huge datasets make epoch validation inefficient.

intents_snips_bigru_val_every_n_batches.json test is
identical to intents_snips_bigru.json except for
"val_every_n_batches" name in "train".

This pull request includes following refactoring
_validate() function is created for validation related ops and used for
validation both after epoch and after batch. Also logging ops
are converted into separate function _log(). Training control variables
 are united into train_monitor dictionary. train_monitor
is initialized in separate function.


PeganovAnton:
I have a question regarding the way 'examples_seen' is computed. When language model is trained batch has shape [batch_size, num_unrollings] and the line (deeppavlov.core.commands.train.py in _train_batches function)
train_monitoring['examples_seen'] += len(x)
does not give correct value for num examples. Is it possible to replace it with
train_monitoring['examples_seen'] += np.array(x).size
?

deeppavlov/core/commands/train.py
```diff
@@ -278,237 +279,253 @@ def _train_batches(model: NNModel, iterator: DataLearningIterator, train_config:
...
-                    model.save()
-                    saved = True
-                else:
-                    patience += 1
-                    log.info('Did not improve on the {} of {}'.format(m_name, best))
 
-                report['impatience'] = patience
-                if train_config['validation_patience'] > 0:
-                    report['patience_limit'] = train_config['validation_patience']
+def _init_train_monitoring(train_config, metrics_functions):
```
yoptar: This could be a class with appropriate methods.

deeppavlov/core/commands/train.py
```diff
@@ -278,237 +279,253 @@ def _train_batches(model: NNModel, iterator: DataLearningIterator, train_config:
...
+        report['loss'] = sum(losses) / len(losses)
+
+    tb_train_writer = train_monitoring['tb_train_writer']
+    if train_monitoring['tb_train_writer'] is not None:
+        for name, score in metrics:
+            metric_sum = tf.Summary(value=[tf.Summary.Value(tag='{}/'.format(tag) + name,
+                                                            simple_value=score), ])
+            tb_train_writer.add_summary(metric_sum, epochs)
+
+        if losses:
```
yurakuratov: this line should be changed to:
`if 'loss' in report:`
like it was done in PR: https://github.com/deepmipt/DeepPavlov/pull/450

deeppavlov/core/commands/train.py
PeganovAnton: Could you give more detail on this idea? Now I can only see that 'improved' key can be replaced with 'improved' method.

PeganovAnton: There can also be clear_after_log method.

PeganovAnton: train_monitoring is a class now

yoptar:
Will be incorporated in #557 solution

PR #460, opened 2018-09-20
Fix/split seed for classification iterator

dilyararimovna:
add split_seed to basic_classification_iterator to fix splitting seed

### DIFF
--- a/deeppavlov/dataset_iterators/basic_classification_iterator.py
+++ b/deeppavlov/dataset_iterators/basic_classification_iterator.py
@@ -36,8 +36,9 @@ class BasicClassificationDatasetIterator(DataLearningIterator):
         field_to_split: name of field (out of ``"train", "valid", "test"``) to split
         split_fields: list of fields (out of ``"train", "valid", "test"``) to which save splitted field
         split_proportions: list of corresponding proportions for splitting
-        seed: random seed
+        seed: random seed for iterating
         shuffle: whether to shuffle examples in batches
+        split_seed: random seed for splitting dataset
         *args: arguments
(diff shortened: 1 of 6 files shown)


deeppavlov/dataset_iterators/basic_classification_iterator.py
```diff
@@ -36,8 +36,9 @@ class BasicClassificationDatasetIterator(DataLearningIterator):
         field_to_split: name of field (out of ``"train", "valid", "test"``) to split
         split_fields: list of fields (out of ``"train", "valid", "test"``) to which save splitted field
         split_proportions: list of corresponding proportions for splitting
-        seed: random seed
+        seed: random seed for iterating
         shuffle: whether to shuffle examples in batches
+        split_seed: random seed for splitting dataset
```
yoptar: I think, it needs to be mentioned that if `split_seed` is `None`, it will be based on the `seed` parameter

deeppavlov/dataset_iterators/basic_classification_iterator.py
```diff
@@ -38,7 +38,7 @@ class BasicClassificationDatasetIterator(DataLearningIterator):
         split_proportions: list of corresponding proportions for splitting
         seed: random seed for iterating
         shuffle: whether to shuffle examples in batches
-        split_seed: random seed for splitting dataset
+        split_seed: random seed for splitting dataset, if `split_seed` is None, division is based on `seed`.
```
yoptar: mono-spaced font is set with double grave
```
``split_seed``
```

PR #471, opened 2018-09-24
fix: remove gpu requirements from ner and squad

mu-arkhipov:
We don't need no
G P U


seliverstov:
We don't need no GPU 
We don't need no thought control 
No dark sarcasm in the classroom 
Teachers leave them CPU alone…

PR #478, opened 2018-09-25
feat: add batch size argument to logit ranker

yurakuratov:


deeppavlov/models/ranking/logit_ranker.py
```diff
@@ -35,8 +35,9 @@ class LogitRanker(Component):
 
     """
 
-    def __init__(self, squad_model, **kwargs):
+    def __init__(self, squad_model, batch_size=50, **kwargs):
```
yoptar: Add typing and update docstrings, please

Issue #112, opened 2018-03-15
Reuse of repeated parameters in config


yurakuratov:
Sometimes duplicated parameters with equal values may appear in config file.

Look at the example below:

"pipe": [
      {
        "name": "squad_preprocessor",
        "context_limit": 400,
        "question_limit": 150,
        "char_limit": 16,
      },
      {
        "name": "squad_vocab_embedder",
        "context_limit": "#squad_prepr.context_limit",
        "question_limit": "#squad_prepr.question_limit",
        "char_limit": "#squad_prepr.context_limit",
      {
        "name": "squad_vocab_embedder",
        "level": "char",
        "context_limit": "#squad_prepr.context_limit",
        "question_limit": "#squad_prepr.question_limit",
        "char_limit": "#squad_prepr.char_limit",
     },
      {
        "name": "squad_model",
        "id": "squad",
        "word_emb": "#vocab_embedder.emb_mat",
        "char_emb": "#char_vocab_embedder.emb_mat",
        "context_limit": "#squad_prepr.context_limit",
        "question_limit": "#squad_prepr.question_limit",
        "char_limit": "#squad_prepr.char_limit",
      }
    ]

It could be more convenient if there was no need to copy _context_limit_, _question_limit_, _char_limit_ and to fill them from something like global parameters scope.

yoptar:
Done in #574

PR #479, opened 2018-09-25
fix: set deeppavlov root from config when installing requirements

yoptar:


yoptar:
does not work until implementation of #112

Issue #520, opened 2018-10-13
Python 3.7.0 support


ololobus:
DeepPavlov has scikit-learn version fixed to v0.19.1, but its c-extensions build fails on python 3.7.0 (at least on macOS), please see [scikit-learn issue](https://github.com/scikit-learn/scikit-learn/issues/11320).

This issue has been fixed in scikit-learn v0.19.2 release, so you have to up at least minor version to enable python 3.7.0 support.

I can try python 3.7.0 compatibility of other packages and prepare a pull-request, if you want.

seliverstov:
Hi, Alexey!
Thanks, It will be great!

ololobus:
Here it is https://github.com/deepmipt/DeepPavlov/pull/521

yoptar:
Waiting on https://github.com/tensorflow/tensorflow/issues/20517

lcswillems:
The issue you are waiting on is now closed. Tensorflow 1.13.1 now supports Python 3.7. Could you update your dependencies?

I would like to use Spacy with your package but I can't currently because of dependencies.

yoptar:
Resolved in #950

PR #521, opened 2018-10-13
Python 3.7 support

ololobus:
Python 3.7 support https://github.com/deepmipt/DeepPavlov/issues/520

Scikit-learn version changed to 0.19.2 in order to address this issue https://github.com/scikit-learn/scikit-learn/issues/11320

pyTelegramBotAPI version changed to the latest to fix test errors, e.g.:
```python
E             File "/Users/alexk/dev/deeppavlov-dev/utils/telegram_utils/telegram_ui.py", line 18, in <module>
E               import telebot
E             File "/Users/alexk/dev/deeppavlov-dev/dp-venv/lib/python3.7/site-packages/telebot/__init__.py", line 1291
E               @util.async()
E                         ^
E           SyntaxError: invalid syntax
```

Pip install and basic functionality works fine with Python 3.7.0 now. Tests do not crash, but take forever on my laptop to complete and some of them are skipped, though the same is with current master with Python 3.6.6.


tests/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pytest>=3.8.2
+pexpect>=4.6.0
```
yoptar: Tests requirements are set as extra dependencies in the `setup.py` file:
https://github.com/deepmipt/DeepPavlov/blob/5b2132709ce51c370629a66b2296033dc5e73d99/setup.py#L57-L60

yoptar:
Thank you for your contribution!
But i thought that TensorFlow does not support Python 3.7 as of yet, so most of our models would still not work with it.

ololobus:
Ok, TF is not directly listed in the requirements.txt, so I thought it is not crytical

aranyasinha:
Tensorflow has supported python3.7 for some time now.

nok:
@aranyasinha is right, now it's possible: https://www.tensorflow.org/install/pip?lang=python3#package-location

yoptar:
There is a PR #861, it is planned to get merged and released this week.

PR #539, opened 2018-10-18
Fix/ner

mu-arkhipov:
Fix some fixes and add other fixes


deeppavlov/models/ner/network.py
```diff
@@ -145,10 +145,10 @@ def __init__(self,
         if net_type == 'rnn':
             if use_cudnn_rnn:
                 if l2_reg > 0:
-                    raise Warning('cuDNN RNN are not l2 regularizable')
+                    log.info('Warning! cuDNN RNN are not l2 regularizable')
```
yoptar:
```suggestion
                    log.warning('cuDNN RNN are not l2 regularizable')
```

PR #550, opened 2018-10-24
fix: UD2.0 sources files not found

seliverstov:


yoptar:
Files themselves have to be restored, not just links

PR #553, opened 2018-10-26
fix: new metrics usage in paramsearch and cross_validation

yoptar:


deeppavlov/core/common/cross_validation.py
```diff
@@ -21,7 +21,8 @@
 
 from deeppavlov.core.common.file import read_json
 from deeppavlov.core.common.log import get_logger
-from deeppavlov.core.commands.train import train_evaluate_model_from_config, get_iterator_from_config, read_data_by_config
+from deeppavlov.core.commands.train import train_evaluate_model_from_config, get_iterator_from_config, \
+    read_data_by_config, get_metrics_names
```
dilyararimovna: you don't have get_metrics_names in `deeppavlov.core.commands.train`

PR #570, opened 2018-11-08
fix: making copy of config in read_data_by_config

dilyararimovna:


deeppavlov/core/commands/train.py
```diff
@@ -19,6 +19,7 @@
 from collections import OrderedDict, namedtuple
 from pathlib import Path
 from typing import List, Tuple, Dict, Union
+from copy import deepcopy
```
yoptar: unused import

dilyararimovna: fixed

PR #582, opened 2018-11-14
Feat/skip downloads when files with the same hashes already exist

yoptar:


deeppavlov/download.py
```diff
@@ -13,17 +13,19 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 """
-
+import shutil
```
nikolay-bushkov: Comment license info with #

deeppavlov/download.py
```diff
@@ -13,17 +13,19 @@
...
 from typing import Union, Optional, Dict, Iterable, Set, Tuple, List
 
+import requests
+
 import deeppavlov
 from deeppavlov.core.commands.utils import get_deeppavlov_root, set_deeppavlov_root, expand_path
 from deeppavlov.core.common.file import read_json
-from deeppavlov.core.data.utils import download, download_decompress, get_all_elems_from_json
+from deeppavlov.core.data.utils import download, download_decompress, get_all_elems_from_json, file_md5
 from deeppavlov.core.common.log import get_logger
```
nikolay-bushkov: PyCharm highlights neither two blank lines, nor one after imports...

utils/prepare/hashes.py
```diff
@@ -0,0 +1,73 @@
+import argparse
```
nikolay-bushkov: So many new lines of code... Isn't it possible to reuse some utils from Linux?
What about https://www.tensorflow.org/api_docs/python/tf/keras/utils/get_file?
Think about maintenance efforts!

deeppavlov/download.py
yoptar: Let's do one

PR #583, opened 2018-11-15
fixed errors and updated docstrings

Guitaricet:
Now this component works does not fail if counter_vocab_path is specified.

### DIFF
--- a/deeppavlov/models/embedders/tfidf_weighted_embedder.py
+++ b/deeppavlov/models/embedders/tfidf_weighted_embedder.py
@@ -30,12 +30,12 @@ log = get_logger(__name__)
 @register('tfidf_weighted')
 class TfidfWeightedEmbedder(Component):
     """
-    The class implements the functional of embedding the sentence \
-        as a weighted by special coefficients average of tokens embeddings. \
+    The class implements the functionality of embedding the sentence \
+        as a weighted average by special coefficients of tokens embeddings. \
         Coefficients can be taken from the given TFIDF-vectorizer in ``vectorizer`` or \
         calculated as TFIDF from counter vocabulary given in ``counter_vocab_path``.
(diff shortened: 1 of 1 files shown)


deeppavlov/models/embedders/tfidf_weighted_embedder.py
```diff
@@ -62,6 +62,18 @@ class TfidfWeightedEmbedder(Component):
         idf_base_count: minimal idf value (less time occured are not counted)
         log_base: logarithm base for TFIDF-coefficient calculation froom counter vocabulary
         min_idf_weight: minimal idf weight
+
+    Examples:
+        >>> from deeppavlov.models.embedders.tfidf_weighted_embedder import TfidfWeightedEmbedder
+        >>> fastTextTfidf([['большой', 'и', 'розовый', 'бегемот']])
```
yoptar:
```suggestion
        >>> from deeppavlov.models.embedders.fasttext_embedder import FasttextEmbedder
```

deeppavlov/models/embedders/tfidf_weighted_embedder.py
```diff
@@ -75,18 +87,12 @@ def __init__(self,
...
         self.dim = self.embedder.dim
         self.mean = mean
         self.pad_zero = pad_zero
-
-        if tokenizer is None:
-            self.tokenizer = self.space_detokenizer
-        else:
-            self.tokenizer = tokenizer
+        self.tokenizer = tokenizer or self.space_detokenizer
+        self.vectorizer = None
```
yoptar:
```suggestion
        self.vectorizer = vectorizer
```
if `vectorizer` is `None`, then it's `None`
And then remove `self.vectorizer = vectorizer` in `elif`

PR #589, opened 2018-11-19
feat: allow validating every n batches

yoptar:


deeppavlov/core/commands/train.py
```diff
@@ -341,7 +342,7 @@ def improved(score, best):
         tb_valid_writer = tf.summary.FileWriter(str(tb_log_dir / 'valid_log'))
 
     # validate first (important if model is pre-trained)
-    if train_config['val_every_n_epochs'] > 0 and epochs % train_config['val_every_n_epochs'] == 0:
+    if train_config['val_every_n_epochs'] > 0 or train_config['val_every_n_batches'] > 0:
         report = _test_model(model, metrics_functions, iterator,
```
vikmary: if `val_everyn_epochs > 1` model will be though tested every epoch.

yoptar: This `if` is not run in any `while` block, it will be executed only once

deeppavlov/core/commands/train.py
```diff
@@ -431,6 +432,46 @@ def improved(score, best):
...
+                                         train_config['batch_size'], 'valid', start_time, train_config['show_examples'])
+                    report['epochs_done'] = epochs
+                    report['batches_seen'] = i
+                    report['train_examples_seen'] = examples
+
+                    metrics = list(report['metrics'].items())
+
+                    if train_config['tensorboard_log_dir'] is not None:
+                        for name, score in metrics:
+                            metric_sum = tf.Summary(value=[tf.Summary.Value(tag='every_n_epochs/' + name,
```
vikmary: Rename path for tf.Summary to `every_n_batches/`

yoptar:
We should refactor `train_batches` a lot, but we don't yet have resources for that.

PR #590, opened 2018-11-19
allow recursive training with a `--recursive` parameter

yoptar:
Tested it with `python -m deeppavlov train --recursive slotfill_dstc2 -d`

### DIFF
--- a/deeppavlov/core/commands/train.py
+++ b/deeppavlov/core/commands/train.py
@@ -30,6 +30,7 @@ from deeppavlov.core.common.params import from_params
 from deeppavlov.core.common.registry import get_model
 from deeppavlov.core.data.data_fitting_iterator import DataFittingIterator
 from deeppavlov.core.data.data_learning_iterator import DataLearningIterator
+from deeppavlov.core.data.utils import get_all_elems_from_json
 from deeppavlov.core.models.estimator import Estimator
 from deeppavlov.core.models.nn_model import NNModel
 from deeppavlov.download import deep_download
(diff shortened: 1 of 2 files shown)


yoptar:
Tested it with `python -m deeppavlov train --recursive slotfill_dstc2 -d`

deeppavlov/core/commands/train.py
```diff
@@ -154,14 +155,20 @@ def get_iterator_from_config(config: dict, data: dict):
...
+                                     to_train=True, to_validate=True, download=False,
+                                     recursive=True) -> Dict[str, Dict[str, float]]:
     """Make training and evaluation of the model described in corresponding configuration file."""
     config = parse_config(config)
 
     if download:
         deep_download(config)
 
+    if to_train and recursive:
+        for subconfig in get_all_elems_from_json(config['chainer'], 'config_path'):
```
my-master: Currently, we can set 'config_path' in any inner attribute of a class in the config.  So I suggest searching 'config_path' recursively in the entire config, not only in first-level elements. Is it possible?
Example from my config (I have to do that to emulate a cycle inside my config):
```
{
        "name": "logit_ranker",
        "squad_model": {
          "config_path": "../deeppavlov/configs/squad/multi_squad_noans.json"
        },
        "top_n": 3,
        "sort_noans": true,
        "in": [
          "chunks",
          "questions"
        ],
        "out": [
          "best_answer",
          "context_indices"
        ]
      },
```

my-master: Oh, I see it's already recursive in `get_all_elems_from_json()`. Alright then!

vikmary:
Checked recursive training for go_bot with trainable slot-filling component.

PR #593, opened 2018-11-20
Yahoo Conversational vs informational model

dilyararimovna:
Add `bigru_with_max_aver_pool_model` to `keras_classification_model`, add config, docs and test for yahoo conversational vs informational model (without opportunity to train, pre-trained model is available only for inference)


deeppavlov/configs/classifiers/intents_snips.json
```diff
@@ -70,8 +70,8 @@
         ],
         "main": true,
         "class_name": "keras_classification_model",
-        "save_path": "{DOWNLOADS_PATH}/classifiers/intents_snips_v7/model",
-        "load_path": "{DOWNLOADS_PATH}/classifiers/intents_snips_v7/model",
+        "save_path": "{MODELS_PATH}/classifiers/intents_snips_v7/model",
+        "load_path": "{MODELS_PATH}/classifiers/intents_snips_v7/model",
```
yoptar: This should be in a separate pull request

deeppavlov/configs/classifiers/yahoo_convers_vs_info.json
```diff
@@ -0,0 +1,174 @@
+{
+  "dataset_iterator": {
```
yoptar: i feel that if there is no reader, there should be no iterator and no `in_y` and `train` blocks

deeppavlov/configs/classifiers/intents_snips.json
dilyararimovna: ok

deeppavlov/configs/classifiers/yahoo_convers_vs_info.json
dilyararimovna: yes. removed

deeppavlov/configs/classifiers/yahoo_convers_vs_info.json
```diff
@@ -0,0 +1,170 @@
...
+        "ref": "my_one_hotter",
+        "in": "y_pred_ids",
+        "out": "y_pred_onehot"
+      }
+    ],
+    "out": [
+      "y_pred_labels"
+    ]
+  },
+  "train": {
```
yoptar: Train block without dataset reader and iterator does nothing

deeppavlov/core/layers/keras_layers.py
```diff
@@ -277,3 +278,35 @@ def compute_output_shape(self, input_shape):
...
+        seq_lengths: tensor of integers of size (batch_size, 2).
+            Each row is a pair (i, length) where length is a number of element of sequences[i] to extract
+
+    Returns:
+        tensor of shape (batch_size, dim)
+    """
+    return Lambda(lambda x: tf.gather_nd(x[0], K.cast(x[1], "int32")))([sequences, seq_lengths])
+
+
+def extracting_part_of_sequences(sequences, seq_lengths):
```
yoptar: `extracting_part_of_sequences` and `masking_sequences` are identical
Also, are you sure, you cannot write them as `Layer` inheritors?

deeppavlov/models/classifiers/keras_classification_model.py
```diff
@@ -162,7 +160,8 @@ def _change_not_fixed_params(self, **kwargs) -> None:
                 self.opt[param] = kwargs.get(param)
         return
 
-    def pad_texts(self, sentences: List[List[np.ndarray]]) -> np.ndarray:
+    def pad_texts(self, sentences: List[List[np.ndarray]],
+                  return_lengths=False) -> Union[np.ndarray, Tuple[np.ndarray, np.ndarray]]:
```
yoptar:
```suggestion
                  return_lengths: bool = False) -> Union[np.ndarray, Tuple[np.ndarray, np.ndarray]]:
```
also, `return_lengths` is not described in docstring

deeppavlov/configs/classifiers/yahoo_convers_vs_info.json
dilyararimovna: I will left this block with only parameter `metrics` to provide user with further understanding of their correct usage

deeppavlov/core/layers/keras_layers.py
dilyararimovna: I actually do not use this functions. I will remove them

deeppavlov/core/layers/keras_layers.py
```diff
@@ -17,6 +17,7 @@
 from keras.layers import Dense, Reshape, Concatenate, Lambda
 from keras.layers.merge import Multiply
 from keras.activations import softmax
+import tensorflow as tf
```
yoptar: `tensorflow` is not used in this file

PR #599, opened 2018-11-22
ODQA with squad noans (best ODQA)

my-master:
- ODQA with multi_squad_noans model
- updated docs

**Warning**
Now ODQA can output empty answer if no other candidates are found.


docs/skills/odqa.rst
```diff
@@ -80,14 +80,17 @@ Comparison
...
-| Model                                               | Dataset        |  Wiki dump            |  F1    |  EM  |
-+-----------------------------------------------------+----------------+-----------------------+--------+------+
-|:config:`DeepPavlov <odqa/en_odqa_infer_wiki.json>`  | SQuAD (dev)    |   enwiki (2018-02-11) |  28.0  | 22.2 |
-+-----------------------------------------------------+----------------+-----------------------+--------+------+
-|`DrQA`_                                              | SQuAD (dev)    |   enwiki (2016-12-21) |   \-   | 27.1 |
-+-----------------------------------------------------+----------------+-----------------------+--------+------+
++-----------------------------------------------------+----------------+---------------------+---------------------+
+|                                                     |                | enwiki (2018-02-11) | enwiki (2016-12-21) |
+|                                                     |                +----------+----------+-----------+---------+
+| Model                                               | Dataset        |  F1      |   EM     |   F1      |   EM    |
```
yoptar: If `Dataset` is the same in all rows, maybe it would be better to move it to the table description?

yoptar:
Results in `features.rst` should also be updated

docs/skills/odqa.rst
my-master: This table will be extended soon with other datasets (TriviaQA and WikiMovies).

PR #623, opened 2018-12-03
Ignore data_type at test if it is empty

Guitaricet:
Useful if you don't have valid/test set and dons't want your training to crash at the end.

### DIFF
--- a/deeppavlov/core/commands/train.py
+++ b/deeppavlov/core/commands/train.py
@@ -251,6 +251,10 @@ def _test_model(model: Chainer, metrics_functions: List[Metric],
 
     expected_outputs = list(set().union(model.out_params, *[m.inputs for m in metrics_functions]))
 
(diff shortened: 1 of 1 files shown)


deeppavlov/core/commands/train.py
```diff
@@ -251,6 +251,10 @@ def _test_model(model: Chainer, metrics_functions: List[Metric],
 
     expected_outputs = list(set().union(model.out_params, *[m.inputs for m in metrics_functions]))
 
+    if not iterator.data[data_type]:
+        log.warning(f'Could not log examples for {data_type}, assuming it\'s empty')
+        return {'metrics': None}
```
yoptar:
```suggestion
        return {'eval_examples_count': 0, 'metrics': None, 'time_spent': str(datetime.timedelta(seconds=0))}
```

PR #645, opened 2018-12-13
feature: new function evaluate_model

dilyararimovna:
add `evaluate_model` function to deeppavlov core

### DIFF
--- a/deeppavlov/__init__.py
+++ b/deeppavlov/__init__.py
@@ -22,13 +22,13 @@ try:
     # noinspection PyUnresolvedReferences
     from .core.commands.train import train_evaluate_model_from_config
     from .download import deep_download
-
-    # TODO: make better and add typing
-    def train_model(config: [str, Path, dict], download: bool = False, recursive: bool = False):
+    from .core.common.chainer import Chainer
+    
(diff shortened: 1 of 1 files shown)


deeppavlov/__init__.py
```diff
@@ -24,9 +24,13 @@
     from .download import deep_download
 
     # TODO: make better and add typing
-    def train_model(config, download=False, recursive=False):
+    def train_model(config: [str, Path, dict], download: bool = False, recursive: bool = False):
```
yoptar: return type?

dilyararimovna: done

deeppavlov/__init__.py
```diff
@@ -22,11 +22,15 @@
     # noinspection PyUnresolvedReferences
     from .core.commands.train import train_evaluate_model_from_config
     from .download import deep_download
-
-    # TODO: make better and add typing
-    def train_model(config, download=False, recursive=False):
+    from .core.common.chainer import Chainer
+    
+    def train_model(config: [str, Path, dict], download: bool = False, recursive: bool = False) -> Chainer:
```
yoptar:
```suggestion
    # TODO: make better
    def train_model(config: [str, Path, dict], download: bool = False, recursive: bool = False) -> Chainer:
```

PR #675, opened 2018-12-28
go_bot compatibility with 0.1.1 release

vikmary:


deeppavlov/models/go_bot/network.py
```diff
@@ -132,6 +136,13 @@ def __init__(self,
                  use_action_mask: bool = False,
                  debug: bool = False,
                  **kwargs):
+        if any(p in network_parameters for p in self.DEPRECATED):
+            raise ConfigError(f"parameters {self.DEPRECATED} are deprecated,"
+                              " for learning rate schedule documentation see"
+                              " deeppavlov.core.models.lr_scheduled_tf_model"
+                              " or read gitub tutorial on super convergence.")
```
yoptar: Usually when something is deprecated, it means that its usage is discouraged but accepted.
Maybe replace this error with a warning log message?

seliverstov: Done.

PR #676, opened 2018-12-29
Feat/variables from environment

### LINKED ISSUE #665: Make ROOT_PATH, MODELS_PATH and DOWNLOADS_PATH environment variables
All config files I've seen so far have the following variables:
```
"ROOT_PATH": "~/.deeppavlov",
"DOWNLOADS_PATH": "{ROOT_PATH}/downloads",
"MODELS_PATH": "{ROOT_PATH}/models"
```
Should we make them environment variables?
This would be better for the following reasons:
1. No need to define the same variables across all configs
1. No need to redefine variables for your system. For example, I do not store source code and downloads/models at same place (cause of their huge size), so I need to change configs for me and change them _back_ to make PR (that is kind of ridiculous). If these variables were in the environment, I'd have to change them only one time - after deeppavlov installation.
1. Make configs working directory independent (no '~/'-paths)

yoptar:
resolves #665

### DIFF
--- a/deeppavlov/core/commands/utils.py
+++ b/deeppavlov/core/commands/utils.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
 from pathlib import Path
 from typing import Union, Dict, TypeVar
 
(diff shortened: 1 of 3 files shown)


my-master:
It's not obvious which variables could be set in the os environment and how it should be done (unfortunately, documentation is not enough). Consider creating `.deeppavlov` with variables as many other libraries have (`.keras`, `.scm_breeze`, `.shutter` and cetera).

PR #692, opened 2019-01-14
fix: partial fit will work without fit_on_batch parameter in config

yoptar:


deeppavlov/core/commands/train.py
```diff
@@ -106,7 +106,10 @@ def fit_chainer(config: dict, iterator: Union[DataLearningIterator, DataFittingI
                 (('fit_on' in component_config) and
                  callable(getattr(component, 'partial_fit', None))):
             component: Estimator
-            targets = component_config.get('fit_on', component_config['fit_on_batch'])
```
my-master: consider `targets = component_config.get('fit_on') or component_config['fit_on_batch']`

PR #697, opened 2019-01-22
Trainer class

yoptar:
- [x] Allow to evaluate on train

- [x] Refactor training process as a class(es)

- [x] Change the way train metrics are calculated (on k new random batches at the moment of logging)

- [x] Allow to choose a trainer class

- [x] Document stuff

- [x] Update existing configs


deeppavlov/core/commands/train.py
```diff
@@ -205,437 +97,44 @@ def train_evaluate_model_from_config(config: [str, Path, dict], iterator=None, *
...
+                log.warning('"to_validate" parameter is deprecated and will be removed in future versions.'
+                            ' Please, use "evaluation_targets" list instead')
+                evaluation_targets = ['test']
+            else:
+                log.warn('Both "evaluation_targets" and "to_validate" parameters are specified.'
+                         ' "to_validate" is deprecated and will be ignored')
 
-            print(json.dumps(report, ensure_ascii=False))
+        res = trainer.evaluate(iterator, evaluation_targets, print_reports=True)
+        trainer.get_chainer().destroy()
```
vikmary: How about creating trainer's own method `destroy()`? For now train.py contains no `Chainer` abstraction.

deeppavlov/core/commands/train.py
```diff
@@ -205,437 +97,44 @@ def train_evaluate_model_from_config(config: [str, Path, dict], iterator=None, *
...
 
-        model.destroy()
+    res = {k: v['metrics'] for k, v in res.items()}
```
vikmary: why returning only metrics?

deeppavlov/core/commands/train.py
yoptar: It can be a little unintuitive.
```python3
my_model = trainer.get_chainer()
trainer.destroy()
result = my_model(batch)
```
This looks like something that should work, but `my_model` is already destroyed

deeppavlov/core/commands/train.py
yoptar: To keep the old API. Cross-validation module depends on it, for example

deeppavlov/core/trainers/trainers.py
```diff
@@ -0,0 +1,447 @@
...
+            report = self.test(data_gen)
+            res[data_type] = report
+            if print_reports:
+                print(json.dumps({data_type: report}, ensure_ascii=False))
+
+        return res
+
+
+@register('nn_trainer')
+class NNTrainer(FitTrainer):
```
vikmary: I see an analogy between (Estimator, NNModel) and (FitTrainer, NNTrainer) abstractions. But NNTrainer inherits from FitTrainer and NNModel doesn't (from Estimator). Do you really want to use FitTrainer, not Trainer name here?

deeppavlov/core/trainers/fit_trainer.py
```diff
@@ -0,0 +1,260 @@
...
+                # noinspection PyPackageRequirements
+                # noinspection PyUnresolvedReferences
+                import tensorflow
+            except ImportError:
+                log.warning('TensorFlow could not be imported, so tensorboard log directory'
+                            f'`{self.tensorboard_log_dir}` will be ignored')
+                self.tensorboard_log_dir = None
+            else:
+                self.tensorboard_log_dir = expand_path(tensorboard_log_dir)
+                self._tf = tensorflow
```
mu-arkhipov: Why not `import tensorflow as tf`?

deeppavlov/core/trainers/fit_trainer.py
```diff
@@ -0,0 +1,260 @@
...
+                in_y = component_config.get('in_y', None)
+                main = component_config.get('main', False)
+                self._chainer.append(component, c_in, c_out, in_y, main)
+        self._built = True
+
+    def _load(self) -> None:
+        if not self._loaded:
+            self._chainer.destroy()
+            self._chainer = build_model({'chainer': self.chainer_config}, load_trained=self._saved)
+            self._loaded = True
```
mu-arkhipov: Should the `self._loaded` be `True` if `self._saved` is `False`?

deeppavlov/core/trainers/fit_trainer.py
```diff
@@ -0,0 +1,260 @@
...
+            show_examples = self.show_examples
+        if metrics is None:
+            metrics = self.metrics
+
+        expected_outputs = list(set().union(self._chainer.out_params, *[m.inputs for m in metrics]))
+
+        outputs = {out: [] for out in expected_outputs}
+        examples = 0
+
+        data = islice(data, self.max_test_batches)
```
mu-arkhipov: Nice slice!

deeppavlov/core/trainers/fit_trainer.py
yoptar: For when we use the trainer only to evaluate a pipeline without training it.

deeppavlov/core/trainers/fit_trainer.py
```diff
@@ -0,0 +1,260 @@
...
+        for x, y_true in data:
+            examples += len(x)
+            y_predicted = list(self._chainer.compute(list(x), list(y_true), targets=expected_outputs))
+            if len(expected_outputs) == 1:
+                y_predicted = [y_predicted]
+            for out, val in zip(outputs.values(), y_predicted):
+                out += list(val)
+
+        if examples == 0:
+            return {'eval_examples_count': 0, 'metrics': None, 'time_spent': str(datetime.timedelta(seconds=0))}
```
mu-arkhipov: Should we raise warning or even error here?

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -0,0 +1,278 @@
...
+
+        report['impatience'] = self.patience
+        if self.validation_patience > 0:
+            report['patience_limit'] = self.validation_patience
+
+        self._chainer.process_event(event_name='after_validation', data=report)
+        report = {'valid': report}
+        print(json.dumps(report, ensure_ascii=False))
+
+    def _log(self, iterator: DataLearningIterator,
```
mu-arkhipov: Event hooks should be here

yoptar: Done

deeppavlov/core/trainers/fit_trainer.py
yoptar: log.warning('Got empty data iterable for scoring')

deeppavlov/core/trainers/fit_trainer.py
yoptar: It will be a pain to manage between different trainers. Can be resolved by using `tensorboardX`, but in another pull-request

deeppavlov/core/trainers/fit_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+                                summary.value.add(tag='partial_fit/' + name, simple_value=score)
+                                writer.add_summary(summary, i)
+                            writer.flush()
+                else:
+                    preprocessed = self._chainer.compute(*iterator.get_instances(), targets=targets)
+                    if len(targets) == 1:
+                        preprocessed = [preprocessed]
+                    result: Optional[Dict[str, Iterable[float]]] = component.fit(*preprocessed)
+
+                    if result is not None and self.tensorboard_log_dir is not None:
```
vikmary: Have you considered making a separate method for wrinting to tensorboar?

yoptar: I did just now, but it looks like this method would have to have a lot of parameters and it would not make the code simpler.

PR #704, opened 2019-02-01
Fix/log

litinsky:
Fixed bugs in DeepPavlov log:
1. DeepPavlov logger configuration overrided all other loggers in the same process. 
2. Turned of DeepPavlov logger propagation to higher level loggers.

### DIFF
--- a/utils/settings/log_config.json
+++ b/utils/settings/log_config.json
@@ -1,37 +1,35 @@
 {
   "version": 1,
-  "disable_existing_loggers": 0,
-  "root":
-  {
-    "level": "DEBUG",
-    "handlers": ["stderr"]
+  "disable_existing_loggers": false,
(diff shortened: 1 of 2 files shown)


utils/settings/log_config.json
```diff
@@ -1,37 +1,35 @@
 {
   "version": 1,
-  "disable_existing_loggers": 0,
-  "root":
-  {
-    "level": "DEBUG",
-    "handlers": ["stderr"]
+  "disable_existing_loggers": false,
+  "loggers": {
+    "_dp": {
```
yoptar:
```suggestion
    "deeppavlov": {
```

my-master: Agree.

my-master: Agree.

PR #720, opened 2019-02-20
feat: add special tokens in spacy_tokenizer

vikmary:
Do not split `stopwords` into subtokens in `StramSpacyTokenizer`.

Example of use:
```python
from deeppavlov.models.tokenizers.spacy_tokenizer import StreamSpacyTokenizer

tok = StreamSpacyTokenizer(alphas_only=False, stopwords=['__PERSON__'])

tok(['__PERSON__ пошел гулять'])
> [['__person__', 'пошел', 'гулять']]
```


seliverstov:
Fixed in #724

Issue #710, opened 2019-02-12
ELMo fine-tuning works incorrectly for different `save_path` and `load_path`


dilyararimovna:
Fine-tuning of ELMo embedding model works incorrectly (does not train at all) if all `save_paths` in config are changed to something different from corresponding `load_paths`.
For example, the config below does not work properly:
```
{
  "dataset_reader": {
    "class_name": "file_paths_reader",
    "data_path": "{DOWNLOADS_PATH}/elmo-lm-ready4fine-example-data/data/",
    "train": "train/*",
    "valid": "heldout/*"
  },
  "dataset_iterator": {
    "class_name": "elmo_file_paths_iterator",
    "seed": 31415,
    "unroll_steps": 20,
    "max_word_length": 50,
    "n_gpus": 1,
    "shuffle": false,
    "bos":"<S>",
    "eos":"</S>",
    "save_path": "{MODELS_PATH}/fine-tuned/vocab.txt",
    "load_path":"{MODELS_PATH}/elmo-lm-ready4fine-tuning-ru-twitter/vocab.txt"
  },
  "chainer": {
    "in": [
      "x_char_ids"
    ],
    "in_y": [
      "y_token_ids"
    ],
    "pipe": [
      {
        "class_name": "elmo_model",
        "options_json_path": "{MODELS_PATH}/elmo-lm-ready4fine-tuning-ru-twitter/options.json",
        "unroll_steps": 20,
        "batch_size": 128,
        "save_path": "{MODELS_PATH}/fine-tuned/saves/model",
        "load_path": "{MODELS_PATH}/elmo-lm-ready4fine-tuning-ru-twitter/saves/model",
        "in": ["x_char_ids", "y_token_ids"],
        "in_y": [],
        "n_gpus": 1,
        "out": ["loss"]
      }
    ],
   "out": [
    "x_char_ids",
    "y_token_ids"
  ]
  },
  "train": {
    "epochs": 20,
    "batch_size": 128,
    "log_every_n_epochs": 1,
    "val_every_n_epochs": 1,
    "validation_patience": 20,
    "metric_optimization": "minimize",
    "metrics": [
      {
        "name": "elmo_loss2ppl",
        "inputs": ["loss"]
      }
    ],
    "tensorboard_log_dir": "{MODELS_PATH}/fine-tuned/logs"
  },
  "metadata": {
    "variables": {
      "ROOT_PATH": "~/.deeppavlov",
[...]

PR #726, opened 2019-02-25
fix: elmo load/save path  #710 and add/fix: add new events for train processes

kudep:
fix elmo load/save path   #710
add new events of nn_trainer class:
- before_validation
- before_train
- before_log


deeppavlov/models/elmo/elmo2tfhub.py
```diff
@@ -198,6 +207,8 @@ def export2hub(weight_file, hub_dir, options):
 
             with tf.Session() as sess:
                 sess.run(tf.global_variables_initializer())
+                if hub_dir.exists():
+                    rmdir(hub_dir)
```
yoptar: https://docs.python.org/3/library/shutil.html#shutil.rmtree

deeppavlov/models/elmo/elmo.py
```diff
@@ -272,6 +273,7 @@ def __init__(self,
         self._build_model(train=False, epoch=load_epoch_num)
 
         self.save()
+        self.load_path = self.save_path
```
yoptar: Please comment on this or use a different variable for reloading model while training

deeppavlov/models/elmo/elmo2tfhub.py
kudep: it's updated

deeppavlov/models/elmo/elmo.py
```diff
@@ -272,6 +273,9 @@ def __init__(self,
         self._build_model(train=False, epoch=load_epoch_num)
 
         self.save()
+        # after building the model and saving to the specified save path
+        # you need to change the way to load the checkpoint
```
yoptar: `you need to change` sounds like a requirement for a user to do something
```suggestion
        # change the way to load intermediate checkpoints
```

PR #736, opened 2019-03-07
feat: reduce search in ignored_words complexity

yoptar:


seliverstov:
The designated problem will be placed in the backlog.

PR #750, opened 2019-03-13
Feat/str rich content

litinsky:
For now DeepPavlov developers expect only str typed utterances and responses to be stored in Agent history. To meet these expectations:
1) RichMessage and RichControl implement __str__ method;
2) Agent utterances and responses are explicitly casted to str when put to agent history.

### DIFF
--- a/deeppavlov/agents/rich_content/default_rich_content.py
+++ b/deeppavlov/agents/rich_content/default_rich_content.py
@@ -30,6 +30,9 @@ class PlainText(RichControl):
         super(PlainText, self).__init__('plain_text')
         self.content: str = text
 
+    def __str__(self, *args, **kwargs) -> str:
+        return self.content
+
     def json(self) -> dict:
         """Returns json compatible state of the PlainText instance.
 
--- a/deeppavlov/agents/rich_content/default_rich_content.py
+++ b/deeppavlov/agents/rich_content/default_rich_content.py
@@ -30,7 +30,7 @@ class PlainText(RichControl):
         super(PlainText, self).__init__('plain_text')
         self.content: str = text
 
-    def __str__(self, *args, **kwargs) -> str:
+    def __str__(self) -> str:
         return self.content
 
     def json(self) -> dict:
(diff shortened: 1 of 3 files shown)


deeppavlov/core/agent/rich_content.py
```diff
@@ -97,6 +100,10 @@ class RichMessage(RichItem):
     def __init__(self) -> None:
         self.controls: list = []
 
+    def __str__(self, *args, **kwargs):
+        result = '\n'.join([str(control) for control in self.controls if str(control)])
+        return result or ''
```
yoptar:
```suggestion
        return result
```
`'\n'.join([])` will already return `''`

deeppavlov/core/agent/rich_content.py
```diff
@@ -97,6 +100,10 @@ class RichMessage(RichItem):
     def __init__(self) -> None:
         self.controls: list = []
 
+    def __str__(self, *args, **kwargs):
+        result = '\n'.join([str(control) for control in self.controls if str(control)])
```
yoptar: If you want to be cool, you can
```suggestion
        result = '\n'.join(filter(bool, map(str, self.controls)))
```

deeppavlov/agents/rich_content/default_rich_content.py
```diff
@@ -30,6 +30,9 @@ def __init__(self, text: str) -> None:
         super(PlainText, self).__init__('plain_text')
         self.content: str = text
 
+    def __str__(self, *args, **kwargs) -> str:
```
yoptar: do you need `*args, **kwargs` here?

deeppavlov/core/agent/rich_content.py
```diff
@@ -97,6 +100,10 @@ class RichMessage(RichItem):
     def __init__(self) -> None:
         self.controls: list = []
 
+    def __str__(self, *args, **kwargs):
```
yoptar: Return type

litinsky: fixed

deeppavlov/agents/rich_content/default_rich_content.py
litinsky: fixed

PR #768, opened 2019-03-18
Add a config name to a download url

vaskonov:
In order to pair the requested file to download with corresponding config file, the config name is added to the requested file url, for instance 

http://files.deeppavlov.ai/deeppavlov_data/ner_conll2003_v4_cpu_compatible.tar.gz?config=ner_conll2003

### DIFF
--- a/deeppavlov/download.py
+++ b/deeppavlov/download.py
@@ -141,8 +141,8 @@ def deep_download(config: Union[str, Path, dict]) -> None:
     downloads = get_configs_downloads(config)
 
     for url, dest_paths in downloads.items():
-        if isinstance(config, Path):
-            url = set_query_parameter(url, 'config', config.stem)
+        if not isinstance(config, dict):
+            url = set_query_parameter(url, 'config', Path(config).stem)
         download_resource(url, dest_paths)
 
 
--- a/deeppavlov/download.py
+++ b/deeppavlov/download.py
@@ -146,7 +146,7 @@ def deep_download(config: Union[str, Path, dict]) -> None:
     downloads = get_configs_downloads(config)
 
     for url, dest_paths in downloads.items():
-        url += "?" + config.stem
+        url += "?config=" + config.stem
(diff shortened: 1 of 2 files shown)


deeppavlov/download.py
```diff
@@ -76,7 +76,12 @@ def get_configs_downloads(config: Optional[Union[str, Path, dict]]=None) -> Dict
 
 
 def check_md5(url: str, dest_paths: List[Path]) -> bool:
-    r = requests.get(url + '.md5')
+    if "?" in url:
+        url_md5 = url.replace("?", '.md5?')
```
yoptar: You can use https://docs.python.org/3.6/library/urllib.parse.html#urllib.parse.urlparse
and https://docs.python.org/3.6/library/collections.html#collections.somenamedtuple._replace

deeppavlov/download.py
```diff
@@ -140,6 +146,7 @@ def deep_download(config: Union[str, Path, dict]) -> None:
     downloads = get_configs_downloads(config)
 
     for url, dest_paths in downloads.items():
+        url += "?config=" + config.stem
```
yoptar: what if the url already has a `?` character?

vaskonov: added a def to treat this case

deeppavlov/download.py
vaskonov: add a def to modify an URL

deeppavlov/download.py
```diff
@@ -140,6 +141,8 @@ def deep_download(config: Union[str, Path, dict]) -> None:
     downloads = get_configs_downloads(config)
 
     for url, dest_paths in downloads.items():
+        if isinstance(config, Path):
+            url = set_query_parameter(url, 'config', config.stem)
```
yoptar: `config` can be a string
```python
        if not isinstance(config, dict):
            url = set_query_parameter(url, 'config', Path(config).stem)
```

PR #724, opened 2019-02-22
[WIP] Add possibility to ignore strings in StreamSpacyTokenizer

my-master:
Refactored #720

PR #781, opened 2019-03-26
fix: update spaCy version

nikolay-bushkov:
Now the library depends on spaCy 2.1
StreamSpacyTokenizer is also refactored, however advanced filtering (#724) is not implemented.


nikolay-bushkov:
Strange problems with docs tests... I think it is not necessary to rebuild the whole thing.

nikolay-bushkov:
https://readthedocs.org/projects/deeppavlov/builds/8817835/

PR #782, opened 2019-03-27
Feat/agent postprocessor

vikmary:
- add abstract Postprocessor to agent
- replace names with user name in DefaultPostprocessor


deeppavlov/core/agent_v2/postprocessor.py
```diff
@@ -0,0 +1,37 @@
+from typing import Sequence
+
+from deeppavlov.models.preprocessors.person_normalizer import PersonNormalizer
+from deeppavlov.models.tokenizers.utils import detokenize
+
+
+class Postprocessor:
```
yoptar:
```suggestion
class Postprocessor(metaclass=ABCMeta):
```

deeppavlov/core/agent_v2/postprocessor.py
```diff
@@ -0,0 +1,37 @@
...
+    def __init__(self):
+        pass
+
+    def __call__(self, responses):
```
yoptar:
```suggestion
    def __call__(self, state: dict):
```

deeppavlov/core/agent_v2/postprocessor.py
```diff
@@ -0,0 +1,37 @@
+from typing import Sequence
+
+from deeppavlov.models.preprocessors.person_normalizer import PersonNormalizer
+from deeppavlov.models.tokenizers.utils import detokenize
+
+
+class Postprocessor:
+    def __init__(self):
+        pass
```
yoptar: Empty `__init__()` can be removed

deeppavlov/core/agent_v2/agent.py
```diff
@@ -31,18 +34,24 @@ def __call__(self, utterances: Sequence[str], user_telegram_ids: Sequence[Hashab
...
+        skill_names, responses, confidences, profiles = self.skill_manager(state)
 
         self._update_profiles(me_users, profiles)
 
-        self.state_manager.add_bot_utterances(me_dialogs, utterances, [datetime.utcnow()] * len(me_dialogs),
+        self.state_manager.add_bot_utterances(me_dialogs, responses, responses,
+                                              [datetime.utcnow()] * len(me_dialogs),
                                               skill_names, confidences)
 
         self._update_annotations(me_dialogs)
```
my-master: It looks like now it makes more sense to annotate 'sent_text', not 'text' in bot utterances.

deeppavlov/core/agent_v2/postprocessor.py
my-master: I don't think Postprocessor should be an abstract class. The basic Postprocessor should be implemented the same way as basic *Preprocessor*, basic *ResponseSelector* and other basic classes. All these basic classes have `rest_caller` param in `__init__()` and not implemented `__call__()`. The idea behind `rest_caller` is that all selectors are separate services. So this postprocessor probably should even have a different name, because it is not 'default'.
And maybe we should even create a single basic class for all these classes.

deeppavlov/core/agent_v2/postprocessor.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def __call__(self, state: dict) -> Sequence[str]:
+        new_responses = []
+        for d_state in state['dialogs']:
+            # get tokens & tags
+            response = d_state['utterances'][-1]
+            ner_annotations = response['annotations']['ner']
+            user_name = d_state['user']['profile']['name']
+            # replace names with user name
+            if ner_annotations and (response['active_skill'] == 'chitchat'):
```
my-master: We should check __any__ ner annotation or only `PERS`?

deeppavlov/core/agent_v2/postprocessor.py
```diff
@@ -0,0 +1,37 @@
...
+class DefaultPostprocessor(Postprocessor):
+    def __init__(self) -> None:
+        self.person_normalizer = PersonNormalizer(per_tag='PER')
+
+    def __call__(self, state: dict) -> Sequence[str]:
+        new_responses = []
+        for d_state in state['dialogs']:
+            # get tokens & tags
+            response = d_state['utterances'][-1]
+            ner_annotations = response['annotations']['ner']
```
my-master: There is a `DialogsParser` class at DeepPavlov, which can be used for parsing states. If there is not enough functionality in this class, we can add more methods to it. 
We should make a server from `Postprocessor` and parse state with `DialogParser`

deeppavlov/core/agent_v2/state_schema.py
```diff
@@ -83,6 +84,7 @@ def to_dict(self):
             'active_skill': self.active_skill,
             'confidence': self.confidence,
             'text': self.text,
+            'sent_text': self.sent_text or self.text,
```
my-master: `or` logic should be included in postprocessor, in this method we only serialize

PR #818, opened 2019-04-25
docs: delete outdated tutorials

yoptar:


dilyararimovna:
Please, restore super convergence tutorial. It is quite informative and useful. Without it nobody will use super convergence because parameters are very difficult.

dilyararimovna:
Please, also restore classification tutorial. It is not outdated and being always updated with all the changes in DeepPavlov.

Issue #784, opened 2019-03-29
Add __repr__() method for Chainer class


yoptar:

yoptar:
Resolved in #819

PR #819, opened 2019-04-26
feat: add __repr__ and __getitem__ methods for chainer

### LINKED ISSUE #785: Add __getitem__() method for Chainer class
`str` argument for getting by id and `int` for getting by index

yoptar:
resolves #785 and #784


yoptar:
also resolves #784

PR #837, opened 2019-05-11
DOC typo fix in ner.rst

kmike:


kmike:
the same typo is fixed in https://github.com/deepmipt/DeepPavlov/pull/833.

PR #861, opened 2019-05-30
Become compatible with python 3.7

### LINKED ISSUE #520: Python 3.7.0 support

yoptar:
resolves #520


Jenkinsfile
```diff
@@ -20,12 +20,17 @@ node('gpu') {
             }
             stage('Tests') {
                 sh """
+                    . /etc/profile
+                    module add cuda/10.0
                     . .venv-$BUILD_NUMBER/bin/activate
-                    flake8 `python -c 'import deeppavlov; print(deeppavlov.__path__[0])'` --count --select=E9,F63,F7,F82 --show-source --statistics
```
vikmary: Is it some kind of syntax check? Why was it removed?

PR #888, opened 2019-06-19
feat: unify default config variables

### LINKED ISSUE #844: Setting `DP_MODELS_PATH` and `DP_DOWNLOADS_PATH` variables doesn't really work
All configs in DeepPavlov library have their own way of what is written in `MODELS_PATH`:
* some configs have just `{ROOT_PATH}/models`
* some configs have one more folder, ex. `{ROOT_PATH}/models/kbqa_mix_lowercase`
* some configs have a lot of additional folders, ex. `{ROOT_PATH}/models/classifiers/insults_kaggle_v3`

This makes pointless setting `DP_MODELS_PATH` environment variable because it will overwrite `MODELS_PATH` and all paths that have additional folders cannot be found then.

I think it could be fixed by standardizing `MODELS_PATH` in all configs as `{ROOT_PATH}/models`, and all additional folders should be in a different variable (` MY_FOLDER = {MODELS_PATH/whatever/folders}`).

yoptar:
Resolves #844

### DIFF
--- a/deeppavlov/configs/classifiers/insults_kaggle.json
+++ b/deeppavlov/configs/classifiers/insults_kaggle.json
@@ -23,8 +23,8 @@
         "fit_on": [
           "y"
         ],
-        "save_path": "{MODELS_PATH}/classes.dict",
-        "load_path": "{MODELS_PATH}/classes.dict",
+        "save_path": "{MODEL_PATH}/classes.dict",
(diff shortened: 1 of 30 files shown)


yoptar:
@dilyararimovna, please check that evolution scripts will not break because of this

PR #895, opened 2019-06-21
BERT tf.keras

nikolay-bushkov:
See also https://github.com/deepmipt/bert/pull/4


deeppavlov/core/layers/bert.py
```diff
@@ -0,0 +1,233 @@
...
+        num_heads: number of attention heads in the Transformer
+        hidden_dropout_prob: dropout probability for the hidden layers
+        attention_probs_dropout_prob: dropout probability of the attention probabilities
+        intermediate_act_fn: the non-linear activation function to apply to the output of the intermediate dense layer
+        layer_norm_epsilon: some small number to avoid zero division during layer normalization
+        trainable: whether the layer's variables should be trainable
+        **kwargs: keyword arguments for base Layer class
+    """
+    def __init__(self,
+                 hidden_size: int = 768,
```
Guitaricet: Remove default values for hidden_size, intermediate_size, num_heads and hidden_dropout_prob.

deeppavlov/core/layers/bert.py
```diff
@@ -0,0 +1,233 @@
...
+                 num_heads: int = 12,
+                 hidden_dropout_prob: float = 0.1,
+                 attention_probs_dropout_prob: float = 0.1,
+                 intermediate_act_fn: Union[str, callable] = gelu,
+                 layer_norm_epsilon: float = 1e-12,
+                 trainable: bool = True,
+                 **kwargs) -> None:
+        super().__init__(**kwargs)
+
+        self.supports_masking = True
```
Guitaricet: Is this a class variable? Should it be changeable after the object is created?

deeppavlov/core/layers/bert.py
```diff
@@ -0,0 +1,233 @@
...
+            return enc
+
+
+class TransformerBlock(tf.keras.layers.Layer):
+    """
+    One block of transformer architecture.
+
+    Args:
+        hidden_size: hidden size of the Transformer (d_model)
+        intermediate_size: the size of the intermediate dense layer
```
Guitaricet: it is not clear how `hidden_size` is different from `intermediate_size`.

deeppavlov/core/layers/embeddings.py
```diff
@@ -0,0 +1,141 @@
...
+from typing import Union, Sequence
+
+import tensorflow as tf
+from tensorflow.python.eager import context
+from tensorflow.python.keras.utils import tf_utils
+from tensorflow.python.framework import ops, tensor_shape
+from tensorflow.python.ops import embedding_ops
+
+
+class AdvancedEmbedding(tf.keras.layers.Layer):
```
Guitaricet: Bad naming. It is by far not clean from it what is the purpose of the class.

deeppavlov/core/layers/embeddings.py
```diff
@@ -0,0 +1,141 @@
...
+        trainable_pos_embedding: whether the positional embedding matrix is trainable
+        **kwargs: keyword arguments for base Layer class
+
+    Note:
+        `token_type_ids` (or `segment_ids`) are calculated on the fly, and user of this layer is responsible
+        for token_type_vocab_size to be enough to deal with the inputs
+    """
+    # TODO: maybe break down this class into three distinct ones
+    def __init__(self,
+                 vocab_size: int = 119547,
```
Guitaricet: Remove defaults for vocab_size, max_len, token_type_vocab_size, sep_token_index and output_dim. They should be specified in BERT model directly.

deeppavlov/core/layers/embeddings.py
Guitaricet: Maybe BERTEmbedder is a better name

Guitaricet: And should it satisfy interface for deeppavlov embedder?

deeppavlov/core/layers/embeddings.py
```diff
@@ -0,0 +1,141 @@
...
+        # tasks that do not have long sequences.
+        self.full_position_emb_table = self.add_weight(shape=(self.max_len, self.output_dim),
+                                                       dtype=tf.float32,
+                                                       initializer=self.embeddings_initializer,
+                                                       trainable=self.trainable_pos_embedding,
+                                                       name='position_embeddings')
+
+    def call(self,
+             token_ids: tf.Tensor,
+             **kwargs) -> tf.Tensor:
```
Guitaricet: It would be better if this method had a docstring that would at least tell us the shapes of output.

deeppavlov/core/layers/bert.py
nikolay-bushkov: d_model looks like quite a common naming for Transformer architecture...
Will it be better to add that intermediate_size is the size of the first layer is pointwise feed-forward block? or use something brand-new `d_inner`? :)

deeppavlov/core/layers/bert.py
nikolay-bushkov: It is a special Keras attribute used to determine whether to compute the output mask (call compute_mask method) while calling the layer instance. You probably can change it, but I don't know common scenarios when you need that.

deeppavlov/core/layers/embeddings.py
nikolay-bushkov: BERTEmbedder was the name of this class some commits ago... I am not sure whether it is only BERT-specific embeddings: without segment embeddings, it looks like common for transformers. The current name is definitely better than EmbeddingPostprocessor (https://github.com/tensorflow/models/blob/master/official/bert/modeling.py#L254).
Breaking down this class into three distinct could simplify naming but I doubt their use cases will go further than usage in the BERT.
Let's wait for some thoughts from other reviewers.

deeppavlov/core/layers/embeddings.py
nikolay-bushkov: Again, we need to think about use cases. If they are 99% of the time limited to RuBERT + MultiBERT-base than these are reasonable defaults.

deeppavlov/core/layers/embeddings.py
nikolay-bushkov: Ok, I will add. However, there is standard comput_output_shape method for shapes :)

deeppavlov/core/layers/attention.py
```diff
@@ -0,0 +1,107 @@
...
+          num_heads: number of attention heads
+          neg_inf: some number close to -infinity to add to logits in order to mask corresponding values
+          attention_probs_dropout_prob: dropout probability of the attention probabilities.
+          trainable: whether the layer's variables should be trainable
+          **kwargs: keyword arguments for base Layer class
+    """
+    def __init__(self,
+                 hidden_size: int = 768,
+                 num_heads: int = 12,
+                 neg_inf: float = -10000.0,
```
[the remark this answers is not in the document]
nikolay-bushkov: I was forced to go to the NumPy docs to discover what you wrote above...
Is it necessary to do for other users?
I made this parameter configurable just because I saw -1e9 in one of the implementations of MultiHeadAttention... But after softmax, you will not see any differences between these two options anyway (I checked).
One of the options is to hardcode this as it is done in the official implementation. What do you think?

deeppavlov/core/layers/attention.py
```diff
@@ -0,0 +1,107 @@
...
+                 **kwargs) -> None:
+        super().__init__(**kwargs)
+
+        self.supports_masking = True
+
+        self.num_heads = num_heads
+        self.hidden_size = hidden_size
+        self.neg_inf = neg_inf
+
+        assert hidden_size % self.num_heads == 0
```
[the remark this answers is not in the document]
nikolay-bushkov: Ok

deeppavlov/core/layers/attention.py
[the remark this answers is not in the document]
Guitaricet: This parameter is useless, so it just spoils the interface.
Yes, we can do hardcode, I just think that the numpy provided one will be better. But I can live with -1e9.

deeppavlov/core/layers/bert.py
Guitaricet: We should at least comment that thing about keras. Just a few words, maybe with link to keras docs.
Also I would recommend to move this parameter to class variables, because it is not object-specific.

deeppavlov/core/layers/bert.py
Guitaricet: Just add to the docstring that `intermediate_size` is input size of the first layer.

deeppavlov/core/layers/embeddings.py
Guitaricet: If you doubt that use cases will go further than usage in the BERT, just name it BERT embedder.

deeppavlov/core/layers/embeddings.py
Guitaricet: Oh, yes. We cannot specify them from BERT model class as I thought before.
Add a comment where are they from.

deeppavlov/core/layers/bert.py
nikolay-bushkov: Ok, I will add some comments. This parameter should stay here because it is by default (from base layer class) set to False, I am just changing its value. 

Thanks for the review!
I will wait for some other reviewers, and then push all the changes including your suggestion.

deeppavlov/core/layers/attention.py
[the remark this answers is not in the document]
yoptar:
```suggestion
        if hidden_size % self.num_heads != 0:
             raise ValueError(f'hidden_size should be divisible by num_heads, got {hidden_size} hidden_size and {num_heads} num_heads instead')
```

yoptar:
We will use https://github.com/huggingface/transformers instead

PR #898, opened 2019-06-25
feat: added /poller endpoint for stand monitoring

IgnatovFedor:
Turned off dialog logging for /poller endpoint and also added filter to ignore poller requests at 'werkzeug' logger.

### DIFF
--- a/deeppavlov/utils/server/server.py
+++ b/deeppavlov/utils/server/server.py
@@ -109,7 +109,7 @@ def interact(model: Chainer, params_names: List[str]) -> Tuple[Response, int]:
 
 
 def test_interact(model: Chainer, params_names: List[str]) -> Tuple[Response, int]:
-    model_args = ["Test string." for _ in params_names]
+    model_args = [["Test string."] for _ in params_names]
     try:
         _ = model(*model_args)
         return Response('["Test passed"]\n'), 200
--- a/deeppavlov/utils/server/server.py
+++ b/deeppavlov/utils/server/server.py
@@ -34,7 +34,7 @@ SERVER_CONFIG_FILENAME = 'server_config.json'
 
 
 class PollerFilter(Filter):
-    pat = re.compile(r'POST\s/\S*poller\s')
+    pat = re.compile(r'POST\s/\S*/poller\s')
     def filter(self, record):
         if PollerFilter.pat.search(record.getMessage()):
             return False
--- a/deeppavlov/utils/server/server.py
+++ b/deeppavlov/utils/server/server.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import re
 import ssl
 from logging import getLogger, Filter
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


deeppavlov/utils/server/server.py
```diff
@@ -100,6 +108,15 @@ def interact(model: Chainer, params_names: List[str]) -> Tuple[Response, int]:
     return jsonify(result), 200
 
 
+def test_interact(model: Chainer, params_names: List[str]) -> Tuple[Response, int]:
+    model_args = ["Test string." for _ in params_names]
```
yoptar: Model takes batches as arguments, so
```suggestion
    model_args = [["Test string."] for _ in params_names]
```
Also, are you sure that all the inputs will always be texts?

deeppavlov/utils/server/server.py
```diff
@@ -154,9 +171,12 @@ def index():
         }
     }
 
-    @app.route(model_endpoint, methods=['POST'])
+    @app.route(model_endpoint, methods=['POST', 'GET'])
```
yoptar: This way the `GET` option appears in Flasgger, it will confuse users. If you're doing `if ... else...` there, you can make a new endpoint without swagger, I feel.

deeppavlov/utils/server/server.py
```diff
@@ -100,6 +108,15 @@ def interact(model: Chainer, params_names: List[str]) -> Tuple[Response, int]:
     return jsonify(result), 200
 
 
+def test_interact(model: Chainer, params_names: List[str]) -> Tuple[Response, int]:
+    model_args = [["Test string."] for _ in params_names]
```
yoptar: Are you sure that all the models will always accept texts for every argument?

yoptar: They could as well work with lists (POS-tags for tokens, for example), so I wouldn't hardcode inputs like this

deeppavlov/utils/server/server.py
```diff
@@ -109,6 +134,12 @@ def start_model_server(model_config, https=False, ssl_key=None, ssl_cert=None, p
     model_endpoint = server_params['model_endpoint']
     model_args_names = server_params['model_args_names']
 
+    if model_endpoint == '/':
+        e = ValueError('"/" endpoint is reserved for Swagger: please provide correct endpoint in model_endpoint'
```
litinsky:
```suggestion
        e = ValueError('"/" endpoint is reserved, please provide correct endpoint in model_endpoint'
```

deeppavlov/utils/server/server.py
```diff
@@ -31,7 +32,18 @@
 
 SERVER_CONFIG_FILENAME = 'server_config.json'
 
+
+class PollerFilter(Filter):
+    pat = re.compile(r'POST\s/\S*/poller\s')
+    def filter(self, record):
+        if PollerFilter.pat.search(record.getMessage()):
+            return False
+        return True
```
yoptar: `return not PollerFilter.pat.search(record.getMessage())`?

deeppavlov/utils/server/server.py
```diff
@@ -31,7 +32,18 @@
 
 SERVER_CONFIG_FILENAME = 'server_config.json'
 
+
+class PollerFilter(Filter):
```
yoptar: A docstring for this class would be great

deeppavlov/utils/server/server.py
IgnatovFedor: done

deeppavlov/utils/server/server.py
IgnatovFedor: Docstring has added

PR #914, opened 2019-07-03
Enable automated testing on Travis CI

cclauss:
Jenkins is just too slow.

The repo would need to be enabled https://travis-ci.com/deepmipt


yoptar:
flake8 check runs on jenkins.
I don't think that we should add travis integration just for it.

PR #932, opened 2019-07-19
Fix/parsing path for evolution

dilyararimovna:


deeppavlov/core/commands/utils.py
```diff
@@ -67,3 +78,10 @@ def import_packages(packages: list) -> None:
     """Import packages from list to execute their code."""
     for package in packages:
         __import__(package)
+
+
+def parse_path_with_config(path: Union[str, Path], config: Union[str, Path, dict]) -> Path:
```
yoptar: maybe something like `parse_value_with_config`?

dilyararimovna: yes, thank you)

PR #935, opened 2019-07-19
Feat/imports in inits

dilyararimovna:
add imports in inits

### DIFF
--- a/deeppavlov/core/data/__init__.py
+++ b/deeppavlov/core/data/__init__.py
@@ -1,0 +1,44 @@
+# Copyright 2019 Neural Networks and Deep Learning lab, MIPT
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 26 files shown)


deeppavlov/core/data/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+try:
+    from .data_fitting_iterator import DataFittingIterator
+except ImportError:
+    pass
```
yoptar: Classes that have to work without additional requirements do not need `try`-`except`

dilyararimovna: I suppose it will be really hard work to check all the classes whether they should not work without additional requirements=(

acriptis: Is it better to log out the warning message in case of ImportError? 

If there is no specific message for particular import exception it may be better to reduce redundancy in code by using only one `try ... except` to wrap all imports:
```
try:
    from .data_learning_iterator import DataLearningIterator
    from .dataset_reader import DatasetReader
    from .simple_vocab import SimpleVocabulary
except ImportError:
    pass
```

yoptar: @acriptis , the idea is to import everything that you can. If you have one try-except block then some imports that could work will not

yoptar:
This way depending on installed additional requirements a user will get different import options and irrelevant error messages.

We will look at this problem later.

PR #950, opened 2019-07-29
Release 0.5.0

yoptar:


deeppavlov/skills/dsl_skill/dsl_skill.py
```diff
@@ -0,0 +1,224 @@
...
+
+class DSLMeta(ABCMeta):
+    """
+    This metaclass is used for creating a skill. Skill is register by its class name in registry.
+
+    Example:
+
+    .. code:: python
+
+            class ExampleSkill(metaclass=DSLMeta):
```
my-master: This example is different from the one given in `dsl_skill.rst` and provides different output. In `dsl_skill.rst` we have an example without `state` argument. If I run the example with `state = "greeting"`, I will get `[['Sorry, I do not understand you'], [0.0]]` output. Without `state = "greeting"` I am getting `[['Hello, my friend!'], [1.0]]` output. Please copypaste either here or to the docs the correct example.

yoptar: Done

PR #962, opened 2019-08-08
feat: Imdb sentiment dataset reader

sgrechanik-h:
This PR implements a dataset reader for the IMDb sentiment classification dataset. It also includes a json configuration for BERT (en, cased) which is mostly the same as the configuration for rusentiment except for the max seq length and batch size (which I set to values such that I don't get out-of-memory on my hardware).

This PR also includes a fix for the `sets_accuracy` metric which should now correctly work for string labels (i.e. wrap them into sets instead converting them to sets). Also I added reporting of cached files in `download_decompress`.

### DIFF
--- a/deeppavlov/core/data/utils.py
+++ b/deeppavlov/core/data/utils.py
@@ -192,6 +192,11 @@ def download_decompress(url: str, download_path: [Path, str], extract_paths=None
         extracted = extracted_path.exists()
         if not extracted and not arch_file_path.exists():
             simple_download(url, arch_file_path)
+        else:
+            if extracted:
+                log.info(f'Found cached and extracted {url} in {extracted_path}')
+            else:
+                log.info(f'Found cached {url} in {arch_file_path}')
     else:
         arch_file_path = download_path / file_name
         simple_download(url, arch_file_path)
--- a/deeppavlov/dataset_readers/imdb_reader.py
+++ b/deeppavlov/dataset_readers/imdb_reader.py
@@ -67,7 +67,7 @@ class ImdbReader(DatasetReader):
                 for filename in labelpath.glob("*.txt"):
                     with filename.open() as f:
                         text = f.read()
-                    data[data_type].append((text, label))
+                    data[data_type].append((text, [label]))
 
             if not data[data_type]:
                 raise RuntimeError(f"Could not load the '{data_type}' dataset, "
--- a/deeppavlov/dataset_readers/imdb_reader.py
+++ b/deeppavlov/dataset_readers/imdb_reader.py
@@ -1,0 +1,76 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from logging import getLogger
+from typing import List, Dict, Any, Optional, Tuple
+from pathlib import Path
+
+from overrides import overrides
+
+from deeppavlov.core.common.registry import register
+from deeppavlov.core.data.dataset_reader import DatasetReader
+from deeppavlov.core.data.utils import download_decompress, mark_done, is_done
+
+log = getLogger(__name__)
+
+
+@register('imdb_reader')
+class ImdbReader(DatasetReader):
+    """This class downloads and reads the IMDb sentiment classification dataset.
+
+    https://ai.stanford.edu/~amaas/data/sentiment/
+
+    Andrew L. Maas, Raymond E. Daly, Peter T. Pham, Dan Huang, Andrew Y. Ng, and Christopher Potts.
+    (2011). Learning Word Vectors for Sentiment Analysis. The 49th Annual Meeting of the Association
+    for Computational Linguistics (ACL 2011).
+    """
+
+    @overrides
+    def read(self, data_path: str, url: Optional[str] = None,
+             *args, **kwargs) -> Dict[str, List[Tuple[Any, Any]]]:
+        """
+        Args:
+            data_path: A path to a folder with dataset files.
+            url: A url to the archive with the dataset to download if the data folder is empty.
+        """
+        data_path = Path(data_path)
+
+        if url is None:
+            url = "http://ai.stanford.edu/~amaas/data/sentiment/aclImdb_v1.tar.gz"
+
+        if not is_done(data_path):
+            log.info('[downloading data from {} to {}]'.format(url, data_path))
+            download_decompress(url, data_path)
+            mark_done(data_path)
+
+        alternative_data_path = data_path / "aclImdb"
+        if alternative_data_path.exists():
+            data_path = alternative_data_path
+
+        data = {"train": [],
+                "test": []}
+        for data_type in data.keys():
+            for label in ["neg", "pos"]:
+                labelpath = Path(data_path) / data_type / label
+                if not labelpath.exists():
+                    raise RuntimeError(f"Cannot load data: {labelpath} does not exist")
+                for filename in labelpath.glob("*.txt"):
+                    with filename.open() as f:
+                        text = f.read()
+                    data[data_type].append((text, label))
+
+            if not data[data_type]:
+                raise RuntimeError(f"Could not load the '{data_type}' dataset, "
+                                   "probably data dirs are empty")
+
+        return data
--- a/deeppavlov/dataset_readers/imdb_reader.py
+++ b/deeppavlov/dataset_readers/imdb_reader.py
@@ -65,7 +65,7 @@ class ImdbReader(DatasetReader):
                 if not labelpath.exists():
                     raise RuntimeError(f"Cannot load data: {labelpath} does not exist")
                 for filename in labelpath.glob("*.txt"):
-                    with filename.open() as f:
+                    with filename.open(encoding='utf-8') as f:
                         text = f.read()
                     data[data_type].append((text, [label]))
 
--- a/deeppavlov/dataset_readers/imdb_reader.py
+++ b/deeppavlov/dataset_readers/imdb_reader.py
@@ -61,7 +61,7 @@ class ImdbReader(DatasetReader):
                 "test": []}
         for data_type in data.keys():
             for label in ["neg", "pos"]:
-                labelpath = Path(data_path) / data_type / label
+                labelpath = data_path / data_type / label
                 if not labelpath.exists():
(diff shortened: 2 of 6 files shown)


dilyararimovna:
You should definitely try conversational English BERT to check whether that classifier will be better or not.

deeppavlov/metrics/accuracy.py
```diff
@@ -49,8 +49,13 @@ def sets_accuracy(y_true: [list, np.ndarray], y_predicted: [list, np.ndarray]) -
     Returns:
         portion of samples with absolutely coincidental sets of predicted values
     """
+    def _to_set(val):
+        if isinstance(val, (list, tuple, set)):
```
yoptar: I would argue that this change is unnecessary, as one can use simple `accuracy` metric for cases where labels are strings. Also there could be other iterable types that are not `list`, `tuple` or `set` that would not be converted to `set` this way but should be.

sgrechanik-h: Yes, I think, you are right, the proper way here is to use `accuracy` instead of `sets_accuracy` (it requires fixing `Proba2Labels` so that it returns just one label when `max_proba=True`).

sgrechanik-h:
@dilyararimovna Conversational BERT seems to be slightly better: ~93.5% accuracy vs ~92.5% accuracy on the original cased english BERT with the same hyperparameters. However, I trained them for about 3 epochs, so maybe I'll try training them for a longer time.

dilyararimovna:
Самое время избавиться от необходимости оборачивать один имеющийся класс в лист классов.
Для этого сделайте следующее:
Замените строку https://github.com/Huawei-MRC-OSI/DeepPavlov/blob/pr-imdb-clean/deeppavlov/dataset_iterators/snips_intents_iterator.py#L31 на
```
result.append((text, intent))
```
Строку https://github.com/Huawei-MRC-OSI/DeepPavlov/blob/pr-imdb-clean/deeppavlov/dataset_readers/basic_classification_reader.py#L37 на 
```
format: str = "csv", class_sep: str = None,
```
Строку  https://github.com/Huawei-MRC-OSI/DeepPavlov/blob/pr-imdb-clean/deeppavlov/dataset_readers/basic_classification_reader.py#L50 на
```
sep (str): delimeter for ``"csv"`` files. Default: None -> only one class per sample
```
Строки https://github.com/Huawei-MRC-OSI/DeepPavlov/blob/pr-imdb-clean/deeppavlov/dataset_readers/basic_classification_reader.py#L90-L93 на
```
                if isinstance(x, list):
                    if class_sep is None:
                        # each sample is a tuple ("text", "label")
                        data[data_type] = [([row[x_] for x_ in x], str(row[y]))
                                           for _, row in df.iterrows()]
                    else:
                        # each sample is a tuple ("text", ["label", "label", ...])
                        data[data_type] = [([row[x_] for x_ in x], str(row[y]).split(class_sep))
                                           for _, row in df.iterrows()]
                else:
                    if class_sep is None:
                        # each sample is a tuple ("text", "label")
                        data[data_type] = [(row[x], str(row[y])) for _, row in df.iterrows()]
                    else:
                        # each sample is a tuple ("text", ["label", "label", ...])
                        data[data_type] = [(row[x], str(row[y]).split(class_sep)) for _, row in df.iterrows()]
```

grwlf:
> It's time to get rid out of necessity to cover one class to list of classes.

My opinion here is that it is better to keep work on sentiment reader conservative. `basic_classification_reader.py` is a common component, many configurations may be affected by changes in it, so I think it is better to do it this kind of work in a separate PR.

Regarding the labels problem I would suggest to make `ImdbReader` to return singleton lists for now to stay compatible with other components.

dilyararimovna:
@grwlf  Then you should revert `proba2labels` component to its previous version. Because the current version returns non-consistent labels.

deeppavlov/dataset_readers/imdb_reader.py
```diff
@@ -0,0 +1,76 @@
...
+
+        data = {"train": [],
+                "test": []}
+        for data_type in data.keys():
+            for label in ["neg", "pos"]:
+                labelpath = Path(data_path) / data_type / label
+                if not labelpath.exists():
+                    raise RuntimeError(f"Cannot load data: {labelpath} does not exist")
+                for filename in labelpath.glob("*.txt"):
+                    with filename.open() as f:
```
yoptar:
```suggestion
                    with filename.open(encoding='utf-8') as f:
```
default encoding can be different on different systems

deeppavlov/dataset_readers/imdb_reader.py
```diff
@@ -0,0 +1,76 @@
...
+
+        alternative_data_path = data_path / "aclImdb"
+        if alternative_data_path.exists():
+            data_path = alternative_data_path
+
+        data = {"train": [],
+                "test": []}
+        for data_type in data.keys():
+            for label in ["neg", "pos"]:
+                labelpath = Path(data_path) / data_type / label
```
yoptar:
```suggestion
                labelpath = data_path / data_type / label
```
`data_path` is already a `Path`

dilyararimovna:
@grwlf Are you planning to release pre-trained classification models?

grwlf:
> @grwlf Are you planning to release pre-trained classification models?

 Do you mean models for this Sentiment task? Not sure about that. I think we could either release the model or provide an instruction on how to fine-tune existing bert. What do you think would be better?
@yoptar ?

yoptar:
> What do you think would be better?
Most of our configs come with pretrained models, so it would be preferable in this case as well.

grwlf:
> > What do you think would be better?
> 
> Most of our configs come with pretrained models, so it would be preferable in this case as well.

OK, we will train the model.

sgrechanik-h:
So, how do we upload the weights to `files.deeppavlov.ai`?

yoptar:
@sgrechanik-h,
You can send me a link to the model on lymar@ipavlov.ai and I'll reply with it's url on files.deeppavlov.ai
Or you can publish a link on some other site in the config.

grwlf:
> > So, how do we upload the weights to `files.deeppavlov.ai`?
> 
> @sgrechanik-h,
> You can send me a link to the model on [lymar@ipavlov.ai](mailto:lymar@ipavlov.ai) and I'll reply with its url on files.deeppavlov.ai
> Or you can publish a link on some other site in the config.

Could you please check this Google.Drive folder? It should contain sentiment_imdb_conv_bert_v0.tar.gz and sentiment_imdb_bert_v0.tar.gz checkpoints.

https://drive.google.com/drive/folders/1fXAR02g-drJKOHhSGND6TLNkhzErz9PE?usp=sharing

deeppavlov/dataset_readers/imdb_reader.py
```diff
@@ -0,0 +1,76 @@
...
+             *args, **kwargs) -> Dict[str, List[Tuple[Any, Any]]]:
+        """
+        Args:
+            data_path: A path to a folder with dataset files.
+            url: A url to the archive with the dataset to download if the data folder is empty.
+        """
+        data_path = Path(data_path)
+
+        if url is None:
+            url = "http://ai.stanford.edu/~amaas/data/sentiment/aclImdb_v1.tar.gz"
```
avmhawk: hidden logic, why should the file be available at this link?

grwlf: Please suggest how to fix.

avmhawk: Now the path must be specified directly in the config. Add "url" to  "dataset_reader": {..}

deeppavlov/dataset_readers/imdb_reader.py
```diff
@@ -0,0 +1,76 @@
...
+
+        if not is_done(data_path):
+            log.info('[downloading data from {} to {}]'.format(url, data_path))
+            download_decompress(url, data_path)
+            mark_done(data_path)
+
+        alternative_data_path = data_path / "aclImdb"
```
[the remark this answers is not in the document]
avmhawk: maybe it's naming problem and you should use "extracted_data_path" for this var.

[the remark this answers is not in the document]
sgrechanik-h: The problem I'm trying to solve here is that the archive contains the `aclImdb` folder where all the data files are located, however `data_path` itself probably ends with `aclImdb`, so the data files are probably somewhere in `~/.deeppavlov/downloads/aclImdb/aclImdb/`. This may create some confusion, and if `data_path` is provided manually, it may be set to `~/.deeppavlov/downloads/aclImdb/aclImdb/` instead of the correct `~/.deeppavlov/downloads/aclImdb` by the user. This logic forgives this mistake.

deeppavlov/dataset_readers/imdb_reader.py
sgrechanik-h: I'm not convinced, many dataset readers of DeepPavlov have their URLs hardcoded, and forcing the user to provide the URL in the config file will harm usability. (However almost all of these harcoded URLs lead to `files.deeppavlov.ai/*`, so the imdb dataset should probably be uploaded to `files.deeppavlov.ai` as well).

deeppavlov/dataset_readers/imdb_reader.py
```diff
@@ -0,0 +1,76 @@
...
+    """This class downloads and reads the IMDb sentiment classification dataset.
+
+    https://ai.stanford.edu/~amaas/data/sentiment/
+
+    Andrew L. Maas, Raymond E. Daly, Peter T. Pham, Dan Huang, Andrew Y. Ng, and Christopher Potts.
+    (2011). Learning Word Vectors for Sentiment Analysis. The 49th Annual Meeting of the Association
+    for Computational Linguistics (ACL 2011).
+    """
+
+    @overrides
```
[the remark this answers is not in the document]
sgrechanik-h: Don't we need it to, I don't know, protect us from misspelling the word `read` or from the future changes of the function's name in the base class?

deeppavlov/core/data/utils.py
```diff
@@ -192,6 +192,11 @@ def download_decompress(url: str, download_path: [Path, str], extract_paths=None
         extracted = extracted_path.exists()
         if not extracted and not arch_file_path.exists():
             simple_download(url, arch_file_path)
+        else:
```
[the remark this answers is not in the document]
sgrechanik-h: Yes, this is much more readable, thanks.

deeppavlov/dataset_readers/imdb_reader.py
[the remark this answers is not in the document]
avmhawk: IMHO, is a redundant solution. we all write code and monitor its consistency, using a third-party little-known library and operations with metaclasses can shoot us in the future.

deeppavlov/dataset_readers/imdb_reader.py
avmhawk: @yoptar what do you think about this? Maybe, i was wrong

deeppavlov/dataset_readers/imdb_reader.py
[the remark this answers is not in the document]
avmhawk: ok, it's not explicit solution, but it works....

PR #963, opened 2019-08-08
 fix: simple_download: make it safer to interrupt downloads

sgrechanik-h:
This PR fixes the `simple_download` function. The problem was that downloads were unsafe to interrupt, because it this cases the partially downloaded file was considered to be fully downloaded. This PR should fix this by downloading into a `*.part` file first and trying to restart the download if such a file exists.

### DIFF
--- a/deeppavlov/core/data/utils.py
+++ b/deeppavlov/core/data/utils.py
@@ -55,29 +55,32 @@ def simple_download(url: str, destination: [Path, str]):
 
     destination = Path(destination)
     destination.parent.mkdir(parents=True, exist_ok=True)
+    temporary = destination.with_suffix(".part")
 
     headers = {'dp-token': get_download_token()}
     r = requests.get(url, stream=True, headers=headers)
     total_length = int(r.headers.get('content-length', 0))
 
     log.info('Downloading from {} to {}'.format(url, destination))
-    with destination.open('wb') as f, tqdm(total=total_length, unit='B', unit_scale=True) as pbar:
-        done = False
-        downloaded = 0
-        while not done:
-            for chunk in r.iter_content(chunk_size=CHUNK):
-                if chunk:  # filter out keep-alive new chunks
-                    downloaded += len(chunk)
-                    pbar.update(len(chunk))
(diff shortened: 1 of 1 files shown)


deeppavlov/core/data/utils.py
```diff
@@ -55,29 +55,32 @@ def simple_download(url: str, destination: [Path, str]):
...
-                if total_length - downloaded != int(r.headers['content-length']):
-                    raise RuntimeError('It looks like the server does not support resuming downloads')
-            else:
-                done = True
+    with temporary.open('ab') as f:
+        downloaded = f.tell()
+        if downloaded != 0:
+            log.warn(f'Found a partial download {temporary}')
+        with tqdm(initial=downloaded, total=total_length, unit='B', unit_scale=True) as pbar:
+            while downloaded < total_length:
```
yoptar: `total_length` can be incorrect in some cases, so if it's `0`, nothing will be downloaded. That's why it was `while not done` before.
It silently breaks on the `ner_conll2003` config, for example.

sgrechanik-h: Oh, ok, I'll fix it.

deeppavlov/core/data/utils.py
```diff
@@ -55,29 +55,40 @@ def simple_download(url: str, destination: [Path, str]):
...
+            while not done:
+                if downloaded != 0:
+                    log.warn(f'Download stopped abruptly, trying to resume from {downloaded} '
+                             f'to reach {total_length}')
+                    headers['Range'] = f'bytes={downloaded}-'
+                    r = requests.get(url, headers=headers, stream=True)
+                    if 'content-length' not in r.headers or \
+                            total_length - downloaded != int(r.headers['content-length']):
+                        raise RuntimeError(f'It looks like the server does not support resuming '
+                                           f'downloads. Please remove {temporary} and try again')
```
yoptar: I've sent a [PR](https://github.com/Huawei-MRC-OSI/DeepPavlov/pull/1) about not asking to remove a temporary file

PR #972, opened 2019-08-16
fix: nn_trainer, don't treat initial validation results as score improvement

grwlf:
Change from this PR allows nn_trainer to save model on existing from `train` function correclty. We assume that model saving shouldn't be ignored if its parameters were changed since the time of previous save. Since `chainer.train_on_batch` changes parameters, we need to mark the model as 'dirty'.

Previous version of DP didn't clear `_saved` flag at all. As a result, the trainer didn't save the last version of model if its `save` method was called at some point during or before training, e.g. in `_validate_first` mode.

### DIFF
--- a/deeppavlov/core/trainers/nn_trainer.py
+++ b/deeppavlov/core/trainers/nn_trainer.py
@@ -73,6 +73,19 @@ class NNTrainer(FitTrainer):
         max_test_batches: maximum batches count for pipeline testing and evaluation, overrides ``log_on_k_batches``,
             ignored if negative (default is ``-1``)
         **kwargs: additional parameters whose names will be logged but otherwise ignored
+
+
+    Trainer saves the model if it sees a progress in scores. The full rules look like following:
+
+    - For the validation savepoint:
+        * 0-th validation (optional). Don't save model, establish a baseline.
+        * 1-th validation.
+             + If we have a baseline, save the model if we see an improvement, don't save otherwise.
+             + If we don't have a baseline, save the model.
+        * 2nd and later validations. Save the model if we see an improvement
+    - For the at-train-exit savepoint:
+        * Save the model if it happened before 1st validation (to capture early training results), don't save otherwise.
+
     """
     def __init__(self, chainer_config: dict, *, batch_size: int = 1,
                  epochs: int = -1,
@@ -98,16 +111,21 @@ class NNTrainer(FitTrainer):
             self.train_metrics = parse_metrics(train_metrics, self._chainer.in_y, self._chainer.out_params)
 
         metric_optimization = metric_optimization.strip().lower()
+        self.score_best = None
+
+        def _improved(op):
+            return lambda score, baseline: False if baseline is None or score is None \
+                                                else op(score,baseline)
+
         if metric_optimization == 'maximize':
-            self.best = float('-inf')
-            self.improved = lambda score: score > self.best
+            self.improved = _improved(lambda a, b: a > b)
         elif metric_optimization == 'minimize':
-            self.best = float('inf')
-            self.improved = lambda score: score < self.best
+            self.improved = _improved(lambda a, b: a < b)
         else:
             raise ConfigError('metric_optimization has to be one of {}'.format(['maximize', 'minimize']))
 
         self.validate_first = validate_first
+        self.validation_number = 0 if validate_first else 1
         self.validation_patience = validation_patience
         self.val_every_n_epochs = val_every_n_epochs
         self.val_every_n_batches = val_every_n_batches
@@ -124,7 +142,7 @@ class NNTrainer(FitTrainer):
         self.patience = 0
         self.last_result = {}
         self.losses = []
-        self.start_time = None
+        self.start_time = None  # type:Optional[float]
 
         if self.tensorboard_log_dir is not None:
             self.tb_train_writer = self._tf.summary.FileWriter(str(self.tensorboard_log_dir / 'train_log'))
@@ -135,7 +153,12 @@ class NNTrainer(FitTrainer):
             raise RuntimeError('Cannot save already finalized chainer')
 
         self._chainer.save()
-        self._saved = True
+
+    def _is_initial_validation(self):
+        return self.validation_number == 0
+    
+    def _is_first_validation(self):
+        return self.validation_number == 1
 
     def _validate(self, iterator: DataLearningIterator,
                   tensorboard_tag: Optional[str] = None, tensorboard_index: Optional[int] = None) -> None:
@@ -159,15 +182,32 @@ class NNTrainer(FitTrainer):
             self.tb_valid_writer.flush()
 
         m_name, score = metrics[0]
-        if self.improved(score):
+
+        # Update the patience
+        if self.score_best is None:
             self.patience = 0
-            log.info('New best {} of {}'.format(m_name, score))
-            self.best = score
+        else:
+            if self.improved(score, self.score_best):
+                self.patience = 0
+            else:
+                self.patience += 1
+
+        # Run the validation model-saving logic
+        if self._is_initial_validation():
+            log.info('Initial best {} of {}'.format(m_name, score))
+            self.score_best = score
+        elif self._is_first_validation() and self.score_best is None:
+            log.info('First best {} of {}'.format(m_name, score))
+            self.score_best = score
+            log.info('Saving model')
+            self.save()
+        elif self.improved(score, self.score_best):
+            log.info('Improved best {} of {}'.format(m_name, score))
+            self.score_best = score
             log.info('Saving model')
             self.save()
         else:
-            self.patience += 1
-            log.info('Did not improve on the {} of {}'.format(m_name, self.best))
+            log.info('Did not improved on the {} of {}'.format(m_name, self.score_best))
 
         report['impatience'] = self.patience
         if self.validation_patience > 0:
@@ -176,6 +216,7 @@ class NNTrainer(FitTrainer):
         self._send_event(event_name='after_validation', data=report)
         report = {'valid': report}
         print(json.dumps(report, ensure_ascii=False))
+        self.validation_number += 1
 
     def _log(self, iterator: DataLearningIterator,
              tensorboard_tag: Optional[str] = None, tensorboard_index: Optional[int] = None) -> None:
@@ -297,5 +338,9 @@ class NNTrainer(FitTrainer):
         else:
             log.warn(f'Using {self.__class__.__name__} for a pipeline without batched training')
 
-        if not self._saved:
+        # Run the at-train-exit model-saving logic
+        if self.validation_number < 1:
+            log.info('Save model to capture early training results')
             self.save()
+
+


yoptar:
The point is to only save the model if the target metric was improved

grwlf:
OK, makes sense. I'm updating the PR, please review the code since I left some FIXMEs. The points of confusion are:
1. `_validate_first` case. if we call `_validate` via `_validate_first` branch then we really compare score with inf and can't be sure if we see an improvement or not. I've included a change fixing this. https://github.com/deepmipt/DeepPavlov/blob/master/deeppavlov/core/trainers/nn_trainer.py#L234
2. `train` may still save the model regardless of score https://github.com/deepmipt/DeepPavlov/blob/master/deeppavlov/core/trainers/nn_trainer.py#L300 
3. Strange logic in base class FitTrainer, where we set `_saved` flag without clear reason.

deeppavlov/core/trainers/fit_trainer.py
```diff
@@ -165,7 +165,7 @@ def get_chainer(self) -> Chainer:
     def train(self, iterator: Union[DataFittingIterator, DataLearningIterator]) -> None:
         """Call :meth:`~fit_chainer` with provided data iterator as an argument"""
         self.fit_chainer(iterator)
-        self._saved = True
+        self._saved = True # FIXME: Should we really set this flag without actually saving the model?
```
yoptar: every fitted component is saved in the `self.fit_chainer(iterator)` method

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -297,5 +302,7 @@ def train(self, iterator: DataLearningIterator) -> None:
         else:
             log.warn(f'Using {self.__class__.__name__} for a pipeline without batched training')
 
+        # FIXME: Should we really save the model without clear improvement in score?
```
yoptar: Sometimes we would want to skip validation step altogether and just train a model for several epochs.

deeppavlov/core/trainers/fit_trainer.py
grwlf: OK

deeppavlov/core/trainers/nn_trainer.py
grwlf: OK, I've included a comment describing this situation.

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -159,15 +159,20 @@ def _validate(self, iterator: DataLearningIterator,
...
+            else:
+                self.patience += 1
+                log.info('Did not improve on the {} of {}'.format(m_name, self.best))
+        else:
             self.patience = 0
-            log.info('New best {} of {}'.format(m_name, score))
+            log.info('Initial best {} of {}'.format(m_name, score))
             self.best = score
-            log.info('Saving model')
-            self.save()
```
yoptar: So if the first validation gives the best metrics then the model will be saved after training with the worse score because `if not self._saved` will activate.
Also, if `self.validate_first` is `False` then after initial validation the model should also be saved.

grwlf: > So if the first validation gives the best metrics then the model will be saved after training with the worse score 
I think we can't call result 'best' before at least 2 measurements. My current intent is to save model at exit before we clearly see an improvement.

> Also, if self.validate_first is False then after initial validation the model should also be saved.
OK, please help me to formulate the rules of desired behavior.

For the validation savepoint:
1) Don't save models during 'validate_first=True' validation (lets call it 0-th validation)
2) Always save the model during 1st validation (to capture early training results).
3) Starting from the 2nd validation, only save the model if the target metric was improved

For the at-train-exit savepont:
3) Save the model if it happened before 1st validation (which is not 'validate_first'), don't save otherwise.

Is it correct?

yoptar: If we're not improving after 0-th validation, we should not save

grwlf: So the rules shold become like follows, please review.

For the validation savepoint:
* 0-th validation (optional). Don't save model, establish a baseline.
* 1-th validation.
   + If we have a baseline and see an improvement, save the model, don't save otherwise.
   + If we don't have a baseline, save the model.
* 2nd and later validations. Save the model if we see an improvement

For the at-train-exit savepont:
* Save the model if it happened before 1st validation (to capture early training results), don't save otherwise.

yoptar: Sounds ok, yes.
But we need to remember to not save at the end if all validations after 0-th did not improve.

grwlf: > But we need to remember to not save at the end if all validations after 0-th did not improve.

Yes, I see, OK, I'll try to implement it.

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -297,5 +351,14 @@ def train(self, iterator: DataLearningIterator) -> None:
...
 
-        if not self._saved:
+        # Run the at-train-exit model-saving logic
+        if self.validation_number < 1:
+            log.info('Save model to capture early training results')
             self.save()
+        else:
+            if self.improved(self.score_best, self.score_baseline):
+                log.info('Save improved model')
+                self.save()
```
yoptar: So if we had multiple validations, best score is better than a baseline, but the latest score is not the best, the latest model will be saved?

grwlf: OK, probably it is incorrect but it looks that I don't get right your last requirement

> But we need to remember to not save at the end if all validations after 0-th did not improve.

Are there any conditions when we want to save models at-exit after 1st validation?

grwlf:
Rebased to 0.6.0.

The current version should follow the below specification, please check. I'm not sure about your last requirement "But we need to remember to not save at the end if all validations after 0-th did not improve".

CC @yoptar 

```
For the validation savepoint:
* 0-th validation (optional). Don't save model, establish a baseline.
* 1-th validation.
   + If we have a baseline, save the model if we see an improvement, don't save otherwise.
   + If we don't have a baseline, save the model.
* 2nd and later validations. Save the model if we see an improvement

For the at-train-exit savepont:
* Save the model if it happened before 1st validation (to capture early training results), don't save otherwise.
```

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -73,6 +73,19 @@ class NNTrainer(FitTrainer):
...
+    Trainer saves the model if it sees a progress in scores. The full rules look like following:
+
+    - For the validation savepoint:
+      * 0-th validation (optional). Don't save model, establish a baseline.
+      * 1-th validation.
+         + If we have a baseline, save the model if we see an improvement, don't save otherwise.
+         + If we don't have a baseline, save the model.
+      * 2nd and later validations. Save the model if we see an improvement
+    - For the at-train-exit savepont:
+      * Save the model if it happened before 1st validation (to capture early training results), don't save otherwise.
```
yoptar: Documentation build fails here because of `Unexpected indentation`

```
    Trainer saves the model if it sees a progress in scores. The full rules look like following:

    - For the validation savepoint:
       * 0-th validation (optional). Don't save model, establish a baseline.
       * 1-th validation.
          + If we have a baseline, save the model if we see an improvement, don't save otherwise.
          + If we don't have a baseline, save the model.
       * 2nd and later validations. Save the model if we see an improvement
    - For the at-train-exit savepoint:
       * Save the model if it happened before 1st validation (to capture early training results), don't save otherwise.
```
works

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -73,6 +73,19 @@ class NNTrainer(FitTrainer):
...
+
+    Trainer saves the model if it sees a progress in scores. The full rules look like following:
+
+    - For the validation savepoint:
+      * 0-th validation (optional). Don't save model, establish a baseline.
+      * 1-th validation.
+         + If we have a baseline, save the model if we see an improvement, don't save otherwise.
+         + If we don't have a baseline, save the model.
+      * 2nd and later validations. Save the model if we see an improvement
+    - For the at-train-exit savepont:
```
yoptar:
```suggestion
    - For the at-train-exit savepoint:
```

deeppavlov/core/trainers/nn_trainer.py
grwlf: Fixed identation

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -98,16 +111,21 @@ def __init__(self, chainer_config: dict, *, batch_size: int = 1,
             self.train_metrics = parse_metrics(train_metrics, self._chainer.in_y, self._chainer.out_params)
 
         metric_optimization = metric_optimization.strip().lower()
+        self.score_best = None
+
+        def _improved(op):
+            return lambda score,baseline: False if baseline is None or score is None \
```
yurakuratov:
```suggestion
            return lambda score, baseline: False if baseline is None or score is None \
```

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -98,16 +111,21 @@ def __init__(self, chainer_config: dict, *, batch_size: int = 1,
...
+                                                else op(score,baseline)
+
         if metric_optimization == 'maximize':
-            self.best = float('-inf')
-            self.improved = lambda score: score > self.best
+            self.improved = _improved(lambda a,b: a>b)
```
yurakuratov:
```suggestion
            self.improved = _improved(lambda a, b: a > b)
```

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -98,16 +111,21 @@ def __init__(self, chainer_config: dict, *, batch_size: int = 1,
...
         elif metric_optimization == 'minimize':
-            self.best = float('inf')
-            self.improved = lambda score: score < self.best
+            self.improved = _improved(lambda a,b: a<b)
```
yurakuratov:
```suggestion
            self.improved = _improved(lambda a, b: a < b)
```

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -124,7 +142,7 @@ def __init__(self, chainer_config: dict, *, batch_size: int = 1,
         self.patience = 0
         self.last_result = {}
         self.losses = []
-        self.start_time = None
+        self.start_time = None # type:Optional[float]
```
yurakuratov:
```suggestion
        self.start_time = None  # type:Optional[float]
```

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -135,7 +153,11 @@ def save(self) -> None:
             raise RuntimeError('Cannot save already finalized chainer')
 
         self._chainer.save()
-        self._saved = True
+
+    def _is_initial_validation(self):
+        return self.validation_number == 0
+    def _is_first_validation(self):
```
yurakuratov:
```suggestion
    
    def _is_first_validation(self):
```

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -159,15 +181,42 @@ def _validate(self, iterator: DataLearningIterator,
...
+                else:
+                    log.info('Did not improved on the {} of {}'.format(m_name, self.score_best))
+        else:
+            if self.improved(score, self.score_best):
+                log.info('Improved best {} of {}'.format(m_name, score))
+                self.score_best = score
+                log.info('Saving model')
+                self.save()
+            else:
+                log.info('Did not improved on the {} of {}'.format(m_name, self.score_best))
```
yurakuratov: can be rewritten as
```
        if self._is_initial_validation():
            log.info('Initial best {} of {}'.format(m_name, score))
            self.score_best = score
        elif self._is_first_validation() and self.score_best is None:
            log.info('Initial best {} of {}'.format(m_name, score))
            self.score_best = score
            log.info('Saving model')
            self.save()
        else:
            if self.improved(score, self.score_best):
                log.info('Improved best {} of {}'.format(m_name, score))
                self.score_best = score
                log.info('Saving model')
                self.save()
            else:
                log.info('Did not improved on the {} of {}'.format(m_name, self.score_best))
```

deeppavlov/core/trainers/nn_trainer.py
grwlf: Really 2 spaces?

yoptar: https://www.python.org/dev/peps/pep-0008/#inline-comments
> Inline comments should be separated by at least two spaces from the statement. They should start with a # and a single space.

deeppavlov/core/trainers/nn_trainer.py
grwlf: Done

PR #973, opened 2019-08-20
Feat/one class is not list

dilyararimovna:
Change `basic_classification_dataset_reader` for the case if `class_sep` is None: for each sample class label is string not list of one element. 
Also changed `proba2labels` component in case `max_proba`: returns one class index not list of one element as previously. 
For all the configs for one-class classification changed `sets_accuracy` to `accuracy`.

### DIFF
--- a/deeppavlov/models/classifiers/proba2labels.py
+++ b/deeppavlov/models/classifiers/proba2labels.py
@@ -70,7 +70,7 @@ class Proba2Labels(Component):
             return [list(np.where(np.array(d) > self.confident_threshold)[0])
                     for d in data]
         elif self.max_proba:
-            return [[np.argmax(d)] for d in data]
+            return [np.argmax(d) for d in data]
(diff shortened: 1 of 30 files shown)


deeppavlov/models/classifiers/proba2labels.py
```diff
@@ -70,7 +70,7 @@ def __call__(self, data: Union[np.ndarray, List[List[float]], List[List[int]]],
             return [list(np.where(np.array(d) > self.confident_threshold)[0])
                     for d in data]
         elif self.max_proba:
-            return [[np.argmax(d)] for d in data]
+            return [np.argmax(d) for d in data]
```
yoptar: there is something weird with this `__call__()` method typing.

dilyararimovna: fixed

PR #996, opened 2019-09-13
fix: fixed wrong model args names

IgnatovFedor:
In some model config model argument name is not in list, that is why rise api instead of making one `question` parameter makes 8 different parameters on every letter of parameter name.


yoptar:
Implemented in #998

PR #999, opened 2019-09-16
Fix parser error

mikenedodai:
When a user dataset after processing in conll 2003 format contains more than 1 `/n`,  an error occurs


deeppavlov/dataset_readers/conll2003_reader.py
```diff
@@ -1,10 +1,11 @@
 from pathlib import Path
+from logging import getLogger
+log = getLogger(__name__)
```
yoptar: Please move `log = getLogger(__name__)` after all the imports, as [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports) states:
> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

deeppavlov/dataset_readers/conll2003_reader.py
```diff
@@ -80,12 +81,20 @@ def parse_ner_file(self, file_name: Path):
                         tags = []
                 else:
                     if self.provide_pos:
-                        token, pos, *_, tag = line.split()
-                        pos_tags.append(pos)
+                        try:
+                            token, pos, *_, tag = line.split()
+                            pos_tags.append(pos)
+                        except:
+                            log.warning('Skip {}, splitted as {}'.format(repr(line), repr(line.split())))
```
yoptar:
```suggestion
                            log.warning('Skip {}, splitted as {}'.format(repr(line), repr(line.split())))
                            continue
```
for both cases. Otherwise `token` and `pos` variables might be repeated from a previous iteration

deeppavlov/dataset_readers/conll2003_reader.py
mikenedodai: thanks for advice

PR #1016, opened 2019-09-26
Docs/installation page

vikmary:
- change location of installation page in docs
- add a few words about installation in quick docs


yoptar:
Merged into dev by hand

PR #1021, opened 2019-10-01
refactor: remove is_str_batch function

yoptar:


vikmary:
`isinstance(np.array(['sdgf])']), str)` returns `False`, but should return true.

yoptar:
Functionality breaks this way

PR #1023, opened 2019-10-01
fix: riseapi now handles List arguments instead of List[str]

IgnatovFedor:


deeppavlov/utils/server/server.py
```diff
@@ -168,29 +176,19 @@ def start_model_server(model_config: Path,
     ssl_config = get_ssl_params(server_params, https, ssl_key=ssl_key, ssl_cert=ssl_cert)
 
     model = build_model(model_config)
-
-    def batch_decorator(cls: MetaModel) -> MetaModel:
-        cls.__annotations__ = {arg_name: List[str] for arg_name in model_args_names}
-        cls.__fields__ = {arg_name: Field(name=arg_name, type_=List[str], class_validators=None,
```
yoptar: Could we not just replace `type_=List[str]` with `type_=list`?

IgnatovFedor: done

deeppavlov/utils/server/server.py
```diff
@@ -149,7 +149,7 @@ def test_interact(model: Chainer, payload: Dict[str, Optional[List]]) -> List[st
         _ = model(*model_args)
         return ["Test passed"]
     except Exception as e:
-        raise HTTPException(status_code=400, detail=e)
+        raise HTTPException(status_code=400, detail=str(e))
```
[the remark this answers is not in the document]
IgnatovFedor: done

PR #1039, opened 2019-10-11
Feat/bert embedder 

vikmary:
- [x] implement bert embedder class and config
- [x] train multilingual and russian sentence representation-based embedders
- [x] put models on share + configs
- [x] move configs for all embedders in `configs/embedder`
- [x] update doc


vikmary:
Example of using bert embedder:
```python
from deeppavlov import build_model, configs

m = build_model(configs.bert_embedder)
 text_embs, word_embs, subword_embs = m(['hi, i want my embedding', 'hi'])
```

PR #1041, opened 2019-10-14
feat: Added ability to use functions in chainer pipe

IgnatovFedor:
Also StrLower class is replaced by str_lower function
Added license info in some files and removed unnecessary imports

### DIFF
--- a/deeppavlov/core/common/chainer.py
+++ b/deeppavlov/core/common/chainer.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import pickle
+from inspect import isclass
 from itertools import islice
 from logging import getLogger
 from types import FunctionType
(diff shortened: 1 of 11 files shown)


yurakuratov:
looks like `str_lower` as function has some problems with serialization, for example:
```
9:17:13  tests/test_quick_start.py::TestQuickStart::test_serialization[ner-ner/ner_dstc2.json-slotfill_dstc2-mode53] FAILED [ 48%]
```

yurakuratov:
one more problem `'function' object has no attribute 'deserialize'`, check this:
https://nbviewer.jupyter.org/gist/yurakuratov/a487b98d5e9c9eec03a5ba5049dc5005

deeppavlov/core/common/chainer.py
```diff
@@ -301,10 +303,12 @@ def destroy(self):
     def serialize(self) -> bytes:
         data = []
         for in_params, out_params, component in self.train_pipe:
-            data.append(component.serialize())
+            serialized = component.serialize() if isclass(component) else None
```
yoptar: `component` is not a class so this will always return `None`

PR #1055, opened 2019-10-27
add f1_samples metric

MrTsepa:
This metric is useful for multilabel classification


yoptar:
Will be added in future versions as a kwarg option for metrics

PR #1060, opened 2019-10-31
fix: allow special characters in sqlite requests values

yoptar:


deeppavlov/core/data/sqlite_database.py
```diff
@@ -156,26 +156,29 @@ def _insert_many(self, data):
...
+        set_fields, where_fields = [], []
+        for k, v in zip(self.keys, record):
+            if k in self.primary_keys:
+                where_fields.append(f"{k}=?")
+                where_values.append(v)
+            else:
+                set_fields.append(f"{k}=?")
+                set_values.append(v)
+        set_expr = ", ".join(set_fields)
+        where_expr = ", ".join(where_fields)
```
yoptar:
```suggestion
        where_expr = " AND ".join(where_fields)
```

PR #1115, opened 2020-01-14
Refactor/tf keras

yoptar:
- replace Keras with tf.keras everywhere
- remove Keras from requirements
- remove test configs for rankings models in favor of using ordinary configs with `pytest_` parameters
- update most of the requirements (including fasttext from pip)
- refactor siamese_iterator


deeppavlov/models/classifiers/keras_classification_model.py
```diff
@@ -529,7 +503,7 @@ def cnn_model_max_and_aver_pool(self, kernel_sizes_cnn: List[int], filters_cnn:
             output_i = Activation('relu')(output_i)
             output_i_0 = GlobalMaxPooling1D()(output_i)
             output_i_1 = GlobalAveragePooling1D()(output_i)
-            output_i = Concatenate()([output_i_0, output_i_1])
+            output_i = concatenate([output_i_0, output_i_1])
```
dilyararimovna: concatenate is a function, not a layer. It's not good because you will have to retrain all pre-trained models

docs/apiref/models/classifiers.rst
```diff
@@ -5,21 +5,9 @@ deeppavlov.models.classifiers
    :members:
 
 .. autoclass:: deeppavlov.models.classifiers.keras_classification_model.KerasClassificationModel
+    :members:
 
     .. automethod:: __call__
-    .. automethod:: pad_texts
-    .. automethod:: train_on_batch
```
dilyararimovna: train_on_batch может оставить?

yoptar: All public methods are included in `:members:`  above.

deeppavlov/models/classifiers/keras_classification_model.py
dilyararimovna: Проверила, compatible, скоры получаются такие же, как должны

deeppavlov/models/morpho_tagger/cells.py
```diff
@@ -12,24 +12,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import keras.activations as kact
-import keras.backend as kb
-import keras.initializers as kinit
-import keras.layers as kl
 import numpy as np
-from keras.engine.topology import InputSpec
+import tensorflow.keras.backend as K
```
AlexeySorokin: One letter variables are misleading, switch back to `kb`.

deeppavlov/models/morpho_tagger/cells.py
```diff
@@ -12,24 +12,24 @@
...
+from tensorflow.keras import activations
```
AlexeySorokin: The name ``activations`` is lowercase and therefore ambigious, it can be easily used in the code with another meaning, switch back to the old notation with ``tensorflow.keras`` instead of ``keras``.

deeppavlov/models/morpho_tagger/cells.py
yoptar: https://keras.io/backend/#using-the-abstract-keras-backend-to-write-new-code
It's how Keras recommends it, so no. We can replace `K` with `tf.keras.backend` and even aspire to replace those with native `tf` calls, but I'm not sure that it's something that we want to do now.

AlexeySorokin: I find no reason to denote ``tf.keras.backend`` with `K`. `K` is a common name for integer parameter, e.g. in loops like for ``for k in range(K)``.

AlexeySorokin: And global one-letter variables are a very bad habit anyway.

deeppavlov/models/morpho_tagger/cells.py
yoptar: refactored

deeppavlov/models/morpho_tagger/cells.py
yoptar: Overruled, as this is a widely accepted notation.

deeppavlov/models/ranking/bilstm_siamese_network.py
```diff
@@ -207,7 +206,8 @@ def _pairwise_distances(self, inputs: List[Tensor]) -> Tensor:
         dot_product = K.dot(embeddings, K.transpose(embeddings))
         square_norm = K.batch_dot(embeddings, embeddings, axes=1)
         distances = K.transpose(square_norm) - 2.0 * dot_product + square_norm
-        distances = K.slice(distances, (0, bs), (bs, bs))
+        # distances = K.slice(distances, (0, bs), (bs, bs))
```
puleon:
```suggestion
```

PR #1158, opened 2020-03-23
use rapidfuzz instead of fuzzywuzzy

maxbachmann:
FuzzyWuzzy is GPLv2 licensed which would force you to licence the whole project under GPLv2. I had the same problem on one of my projects and so I wrote [rapidfuzz](https://github.com/rhasspy/rapidfuzz) which is implementing the same algorithm but is based on a version of fuzzywuzzy that was MIT Licensed and is therefor MIT Licensed aswell, so it can be used in here without forcing a License change. As a nice bonus it is fully implemented in C++ and comes with a few Algorithmic improvements making it between 5 and 100 times faster than FuzzyWuzzy.

### DIFF
--- a/deeppavlov/requirements/spelling.txt
+++ b/deeppavlov/requirements/spelling.txt
@@ -1,4 +1,3 @@
 lxml==4.4.2
-python-Levenshtein==0.12.0
 sortedcontainers==2.1.0
 sacremoses==0.0.35
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 aio-pika==6.4.1
 Cython==0.29.14
 fastapi==0.47.1
-rapidfuzzy==0.2.1
+rapidfuzz==0.2.1
 h5py==2.10.0
 nltk==3.4.5
 numpy==1.18.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 aio-pika==6.4.1
 Cython==0.29.14
 fastapi==0.47.1
-fuzzywuzzy==0.17.0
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
 aio-pika==6.4.1
 Cython==0.29.14
 fastapi==0.47.1
-fuzzywuzzy==0.17.0
+rapidfuzzy==0.2.1
```
yoptar:
```suggestion
rapidfuzz==0.2.1
```

deeppavlov/requirements/spelling.txt
```diff
@@ -1,4 +1,3 @@
 lxml==4.4.2
-python-Levenshtein==0.12.0
```
yoptar: Why remove it?

maxbachmann: python-Levenshtein is installed to speedup fuzzywuzzy and not required for rapidfuzz anymore. Beside this it is GPLv2 Licensed aswell

yoptar: Looks like it's not used anywhere and definitely not in any of the current spelling correction algorithms, so it should be removed from here, thank you :slightly_smiling_face:

maxbachmann: it is used by fuzzywuzzy to speed up the ratio calculation.

yoptar:
Thank you for this PR, we'll most likely use `rapidfuzz` in future DeepPavlov versions, but I'll close this PR for now just to not run tests on it every time.

PR #1159, opened 2020-03-23
tests: test python inference instead of `interact` mode

yoptar:
We can stop implement workarounds for strings as inputs for some models if we do not guarantee that the `interact` mode will work.

### DIFF
--- a/tests/test_quick_start.py
+++ b/tests/test_quick_start.py
@@ -355,7 +355,7 @@ def _serialize(config):
 
 def _infer(config, inputs, download=False):
     chainer = build_model(config, download=download)
-    return chainer(*inputs)
+    return chainer(*inputs) if inputs else []
 
 
 def _deserialize(config, raw_bytes, examples):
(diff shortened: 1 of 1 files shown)


tests/test_quick_start.py
```diff
@@ -368,30 +373,19 @@ def _deserialize(config, raw_bytes, examples):
...
-                        f"Error in interacting with {model_directory} ({config_path}): {query}"
-
-            p.expect("::")
-            p.sendline("quit")
-            p.readlines()
-            if p.wait() != 0:
-                raise RuntimeError('Error in quitting from deep.py: \n{}'.format(logfile.getvalue().decode()))
-        except pexpect.exceptions.EOF:
-            raise RuntimeError('Got unexpected EOF: \n{}'.format(logfile.getvalue().decode()))
+    def interact(config_path, qr_list=None, check_outputs=True):
```
oserikov: Don't we have to stop calling the method `interact` when migrating from *interact* to *infer* behavior ?

PR #1172, opened 2020-04-08
Fix config for trans_en_mbert

gmorgachev:
Fix `KeyError: 'PORTION'` during install trans_en_mbert


vaskonov:
Already resolved. Thank you.

Issue #1189, opened 2020-04-27
 ImportError: cannot import name 'build_model'


SahithiParsi:
Hi team,

I have installed DeepPavlov version 0.1.0 but unable to import build_model. PFA screenshot for details.
Configuration: Windows 10

![image](https://user-images.githubusercontent.com/16195067/80368774-e2da5a00-88aa-11ea-9626-fd8703297131.png)

IgnatovFedor:
Hi, @SahithiParsi 
Please, send result of the following code:
import deeppavlov
print(deeppavlov.__version__, deeppavlov.__file__)

SahithiParsi:
![image](https://user-images.githubusercontent.com/16195067/80372070-8b3eed00-88b0-11ea-923c-c7658aaafe34.png)

IgnatovFedor:
@SahithiParsi, could you also send result of `deeppavlov.__dir__()`?

IgnatovFedor:
@SahithiParsi, we found out what caused this error. Some libraries in your virtual environment was installed with errors, so import error was raised when you imported deeppavlov. Original import error was muted by our library, that's why you can import deeppavlov without any errors. Error occured after initialization of deeppavlov.configs, that's why `ImportError: cannot import name "build_model"` error was raised but configs was imported successfully, although they are declared in one file.
We prepared fix for error hiding issue in DeepPavlov. Now you can edit `C:\\Users\sahithi\Anaconda3\lib\site-packages\deeppavlov\__init__.py` like in this commit: https://github.com/deepmipt/DeepPavlov/pull/1200/commits/98cc69f9cbc44ea10561a5a1a7bb019f4aadf4bd (although it’s better to use `sys.stderr` instead of `sys.stdout`). After these changes you can see which library is imported incorrectly.

Issue #1067, opened 2019-11-07
Broken install on alpine docker


DaniloZZZ:
Using this Dockerfile, deeppavlov installes, but the installation seems to be broken

```
from python:3.7-alpine
run apk add g++ python3-dev git make cmake

# Use wheels, because building scipy from source is a pain. https://github.com/pypa/pip/issues/3969
run echo "manylinux1_compatible = True" >> /usr/local/lib/python3.7/_manylinux.py
run pip3 install deeppavlov
run python3 -c "from deeppavlov import train_model"

```
I get error on last command
`ImportError: cannot import name 'train_model' from 'deeppavlov' (/usr/local/lib/python3.7/site-packages/deeppavlov/__init__.py)`

litinsky:
Hi! Thanks for the info. Yes, we discovered problems with DeepPavlov on Alpine and started investigating solution. If you need to build some solution with DeepPavlov on Docker, you can try one of our pre-buillt Docker images: https://hub.docker.com/r/deeppavlov/base-cpu

praneethgb:
Hi @litinsky, Are Deeppavlov on Alpine problems resolved now?.

IgnatovFedor:
Hi, @DaniloZZZ 
The problem was with numpy in Alpine image: `Error loading shared library ld-linux-x86-64.so.2: No such file or directory (needed by /usr/local/lib/python3.7/site-packages/numpy/core/_multiarray_umath.cpython-37m-x86_64-linux-gnu.so)`. This error was muted by DeepPavlov, that's why only `cannot import name 'train_model' from 'deeppavlov'` occured. We have prepared a bug concealment fix for DeepPavlov, but can't fix usage of numpy installed via pip in Alpine.

PR #1200, opened 2020-05-06
feat: added traceback echo if importerror occurs

IgnatovFedor:
Should help to solve issues like #1067 and #1189

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,18 +12,28 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 
+import importlib.util
+from pathlib import Path
+
 import sphinx_rtd_theme
 
-import deeppavlov
+# Import __version__ from deeppavlov may cause issues in the process of
+# development: __version__ is taken from _meta.py which is updated only after
+# running setup.py. A developer may be puzzled why the version change is not
+# reflected in the documentation (if he didn't run installation after that).
+# We can't use relative import from setup.py, therefore forced to use importlib.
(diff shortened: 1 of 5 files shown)


docs/conf.py
```diff
@@ -12,18 +12,28 @@
...
+setup = importlib.util.module_from_spec(spec)
+spec.loader.exec_module(setup)
 
 # -- Project information -----------------------------------------------------
 
 project = 'DeepPavlov'
-copyright = '2018, ' + deeppavlov.__author__
-author = deeppavlov.__author__
+copyright = f'2018, {setup.__author__}'
+author = setup.__author__
```
yoptar, line 33: This doesn't feel right...

setup.py
```diff
@@ -14,11 +14,30 @@
...
+    meta_path = os.path.join(__location__, 'deeppavlov', '_meta.py')
+    with open(meta_path, 'w') as meta:
+        meta.writelines([
+            f'__version__ = {repr(__version__)}\n',
+            f'__author__ = {repr(__author__)}\n',
+            f'__description__ = {repr(__description__)}\n',
+            f'__keywords__ = {repr(__keywords__)}\n',
+            f'__license__ = {repr(__license__)}\n',
+            f'__email__ = {repr(__email__)}\n'
+        ])
```
yoptar, line 38: We can do a `_meta.py` file and then do:
```python3
with open(meta_path) as meta:
    exec(meta)
```
to get all the variables without importing the `deeppavlov` package.
Would it be better?

docs/conf.py
IgnatovFedor: Only change 2018 -> 2020?

yoptar: Importing `setup` in docs doesn't feel right for some reason (=

PR #1215, opened 2020-05-13
fix: allow excluding load_path argument for bert_classifier

### LINKED ISSUE #1160: Dont load previous model in bert_classifier
I found that on each run my model shows better results. It was suspicious. I realized it uses the same trained model. I tried to remove "load_path" from config file in "train" section and it raised an exception File "....\Anaconda3\lib\site-packages\deeppavlov\models\bert\bert_classifier.py", line 98, in __init__
    and not tf.train.checkpoint_exists(str(self.load_path.resolve())):
AttributeError: 'NoneType' object has no attribute 'resolve'
What is good way to train a fresh model everytime and not to delete files manually before each training?

yoptar:
fixes #1160


yurakuratov:
The same logic might be added to other bert-based models.

PR #1252, opened 2020-06-17
fix: fix typing

PeganovAnton:
If GRU cell is used the hidden size is `int`.

### DIFF
--- a/deeppavlov/core/layers/tf_layers.py
+++ b/deeppavlov/core/layers/tf_layers.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from logging import getLogger
-from typing import List
+from typing import List, Union
 
 import numpy as np
 import tensorflow as tf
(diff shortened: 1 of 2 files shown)


deeppavlov/core/layers/tf_layers.py
```diff
@@ -115,7 +115,7 @@ def dense_convolutional_network(units: tf.Tensor,
 
 
 def bi_rnn(units: tf.Tensor,
-           n_hidden: List,
+           n_hidden: Union[List, int],
```
yoptar, line 118: Are you sure it can be an `int`? This typing will contradict the doctstring.
Also, there's a `n_hidden_list[-1]` reference in the `Returns` block of the docstring

PeganovAnton: According to tf docs parameter `n_hidden` of  `GRUCell` constructor is `int`. I fixed docstring mentions of `n_hidden`.

PR #1263, opened 2020-06-30
improve json logging

dimakarp1996:
Fixed bug when numpy array, which was output from the model, was not shown in the evaluator. This fix is related to this issue https://github.com/deepmipt/DeepPavlov/issues/9331


dimakarp1996:
Is anything lacking here?

oserikov:
just passing by.. but

> Is anything lacking here?

link to the issue #9331 leads to nowhere 0_o

dimakarp1996:
Long story short, it makes evaluator log numpy arrays as well, instead of crashing when np array is the output

PR #1287, opened 2020-07-31
Rewritten text from start to "Not so minimal demo"

danielkornev:
This is Part 1 of the review.

For the Part 2 of the review, it is advised for the author of this document (iPython notebook) to re-write the "Not so minimal demo" in the following way:

1. Paragraph explaining the domain area (restaurants, intents, etc.)
2. Show listings of the three files (stories.md, nlu.md, domain.yml)
3. Give an optional (expand/hide) area with explanation that these files have been automatically generated from the DSTC-2 file format
4. Explain that slot filler is configured and trained separately from the go-bot itself
5. Complete this section in the manner similar to one used in "Basic Demo"

Once this is done, I'll be happy to provide my second Part of the review.


oserikov:
thx, working on it

PR #1351, opened 2020-11-25
Feat/hf classifier autoconfig

vaskonov:


deeppavlov/configs/classifiers/insults_kaggle_bert_torch.json
```diff
@@ -19,7 +19,7 @@
     "pipe": [
       {
         "class_name": "torch_bert_preprocessor",
```
yurakuratov, line 21: torch_bert_preprocessor -> torch_transformers_preprocessor

deeppavlov/configs/classifiers/insults_kaggle_bert_torch.json
```diff
@@ -59,7 +59,7 @@
         "class_name": "torch_bert_classifier",
```
yurakuratov, line 59: torch_bert_preprocessor -> torch_transformers_preprocessor

yurakuratov (commented):
I think that it is needed to check which pre-trained models from this list https://huggingface.co/transformers/pretrained_models.html
are working with DeepPavlov classification pipeline and at least specify supported and unsupported models.

UPD: I have tried to use DialoGPT-small and got
```
ValueError: Unrecognized configuration class <class 'transformers.configuration_gpt2.GPT2Config'> for this kind of AutoModel: AutoModelForSequenceClassification.
Model type should be one of DistilBertConfig, AlbertConfig, CamembertConfig, XLMRobertaConfig, BartConfig, RobertaConfig, BertConfig, XLNetConfig, FlaubertConfig, XLMConfig
```
So, this is that list of models that could be used for classification.

deeppavlov/configs/classifiers/insults_kaggle_bert_torch.json
vaskonov: maybe a single better than plural?
torch_transformer_preprocessor

vaskonov: @IgnatovFedor we need your opinion here

yurakuratov: When I wrote `transformers` I thought about HuggingFace Transformers, not about NN architecture. Don't know what is better here.

IgnatovFedor: torch_transformers_preprocessor, agree with @dilyararimovna

IgnatovFedor (approved):

PR #1481, opened 2021-08-25
Deeppavlov fixes

dimakarp1996:
1) Added tqdm in trainer
2) Added handling separate-class labels as float labels if specified
3) Addled handling quotes while reading dataset if specified

### DIFF
--- a/deeppavlov/core/data/utils.py
+++ b/deeppavlov/core/data/utils.py
@@ -465,7 +465,7 @@ def flatten_str_batch(batch: Union[str, Iterable]) -> Union[list, chain]:
         ['a', 'b', 'c', 'd']
 
     """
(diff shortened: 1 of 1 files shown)


deeppavlov/core/data/utils.py
```diff
@@ -465,7 +465,7 @@ def flatten_str_batch(batch: Union[str, Iterable]) -> Union[list, chain]:
         ['a', 'b', 'c', 'd']
 
     """
-    if isinstance(batch, str):
+    if isinstance(batch, str) or isinstance(batch, str) or isinstance(batch, int) or isinstance(batch, float):
```
vaskonov, line 468: Twice `str` is mentioned.
Consider replacing `isinstance` with `type`.

PR #1487, opened 2021-09-16
Better logging message if wrong pipeline

dimakarp1996:


dimakarp1996:
Will be a part of PAL-BERT PR

PR #1497, opened 2021-10-10
Fix NumpyArrayEncoder

dimakarp1996:
This PR solves the problem with class NumpyArrayEncoder,that does not correctly handle all types of possible inputs.


IgnatovFedor (changes requested):
Add description what problem this PR solves

PR #1501, opened 2021-11-04
8bit optimizers

dimakarp1996:
Added 8bit optimizers from https://github.com/facebookresearch/bitsandbytes to the DeepPavlov library. To be tested yet.


IgnatovFedor (changes requested):
What are the results of tests?

PR #1503, opened 2021-11-11
add roberta support

BunnyNoBugs:
Add RoBERTa support in `squad_torch_bert` config.

The changes concern the shape of `token_type_ids` specific for RoBERTa

### DIFF
--- a/deeppavlov/models/torch_bert/torch_transformers_squad.py
+++ b/deeppavlov/models/torch_bert/torch_transformers_squad.py
@@ -128,7 +128,7 @@ class TorchTransformersSquad(TorchModel):
         y_end = [x[0] for x in y_end]
         b_y_st = torch.from_numpy(np.array(y_st)).to(self.device)
         b_y_end = torch.from_numpy(np.array(y_end)).to(self.device)
-        
+       
         input_ = {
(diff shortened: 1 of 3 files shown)


deeppavlov/models/torch_bert/torch_transformers_squad.py
```diff
@@ -184,7 +184,9 @@ def __call__(self, features: List[InputFeatures]) -> Tuple[List[int], List[int],
         b_input_ids = torch.cat(input_ids, dim=0).to(self.device)
         b_input_masks = torch.cat(input_masks, dim=0).to(self.device)
         b_input_type_ids = torch.cat(input_type_ids, dim=0).to(self.device)
-        
+        if self.pretrained_bert == 'roberta-base':
```
vaskonov, line 187: it seems like we should generalize this condition (not only to roberta-base)

BunnyNoBugs: updated

BunnyNoBugs:
Now distilbert, roberta, xlm, bart and longformer-like models can be trained correctly (confirmed on a GPU)

PR #1552, opened 2022-04-14
feat: chainer profiling

Ihab-Asaad:


deeppavlov/core/commands/infer.py
```diff
@@ -43,9 +42,7 @@ def build_model(config: Union[str, Path, dict], mode: str = 'infer',
     import_packages(config.get('metadata', {}).get('imports', []))
 
     model_config = config['chainer']
-
-    model = Chainer(model_config['in'], model_config['out'], model_config.get('in_y'))
-
+    model = Chainer(model_config['in'], model_config['out'], model_config.get('in_y'), model_config.get('buckets'))
```
IgnatovFedor, line 45: get buckets and histogram name from model_config['metadata']['profiling']
If there is no such field, don't initialize histogram, if there is such dict with parameters, initialize histogram here.
Initialized histogram should be passed to Chainer instead of buckets.

IgnatovFedor (changes requested):

deeppavlov/core/trainers/nn_trainer.py
```diff
@@ -293,6 +294,9 @@ def train_on_batches(self, iterator: DataLearningIterator) -> None:
 
                 if self.log_every_n_batches > 0 and self.train_batches_seen % self.log_every_n_batches == 0:
                     self._log(iterator, tensorboard_tag='every_n_batches', tensorboard_index=self.train_batches_seen)
+                    if self._chainer.hist_name is not None:
+                        self._chainer.print_hist()
+
```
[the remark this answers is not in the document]
Ihab-Asaad: I added this lines in case the user wants to see the statistics, otherwise how can we use this addition?

deeppavlov/core/commands/infer.py
Ihab-Asaad: so statistics will be valid in inference mode only

Ihab-Asaad: initialize hist in case "profiling" exist without "buckets" ?

PR #1555, opened 2022-04-18
remove russian comments from levenshtein_searcher.py

Ihab-Asaad:


IgnatovFedor (changes requested):
Format translated/updated docstrings according [Google Style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) and replace ``[to_]make_cashed`` with ``[to_]make_cached``.
And remove unnecessary newlines in docstrings according to line limit 120 characters

PR #1556, opened 2022-04-18
remove russian comments from levenshtein_searcher.py

Ihab-Asaad:


IgnatovFedor:
Duplicates #1555
`remove/russian_comments` could be merged locally to `issues-1418/1420` branch allowing not to open 1555, but after openning 1555 there is no need in such merge

PR #1574, opened 2022-06-23
remove: TensorFlow

IgnatovFedor:
- Removed all tf-based components.
- TensorFlow removed from the documentation.
- Default train class replaced with `torch_trainer`


vaskonov (approved):
Approved

PR #1579, opened 2022-07-11
Multitask transformer-agnostic model

dimakarp1996:
Added multitask transformer-agnostic model. Supported kinds of tasks - classification, regression, NER, multiple choice. 

FIle multitask_example.json is an example of using multitask with all these kinds of tasks.
FIle config_glue.json is the example of GLUE config. All configs are being trained and evaluated in the same way as original DeepPavlov configs.

File proba2labels.py is modified so as we can use it with many inputs rather than with one, thus simplifying config.
Note that, to increase multitask performance, it is better to have only 1 task in every batch ( just like author of http://github.com/namisan/mt-dnn did). However, while iterating we need to have samples from every task on every step. So in that code Nones are passed where we have no samples to solve this problem. Correct handling of these Nones required modification in core.

Multitask model also supports single-task mode.

As the model is transformer-agnostic, it supports different kinds of transformers.


dimakarp1996:
Тесты не прошли только из-за того, что не скачался файл vocab.txt в ner_rured с удаленного сервера huggingface, что не связано с внесенными мной изменениями.

deeppavlov/configs/multitask/config_glue.json
```diff
@@ -0,0 +1,360 @@
...
+      "log_every_n_epochs":1,
+      "show_examples":false,
+      "evaluation_targets":[
+         "valid"
+      ],
+      "class_name":"torch_trainer"
+   },
+   "metadata":{
+      "variables":{
+         "ROOT_PATH":"/cephfs/home/karpov.d",
```
vaskonov, line 352: !

utils/glue_inference/evaluate_glue.py
```diff
@@ -0,0 +1,192 @@
...
+    data = defaultdict(lambda: defaultdict(int))
+    true = 0
+    tot = 0
+    for ii, jj in zip(a, b):
+	data[ii][jj] += 1
+	if ii == jj:
+	    true += 1
+	tot += 1
+    if true == 0:
+	breakpoint()
```
vaskonov, line 42: !

utils/glue_inference/evaluate_glue.py
```diff
@@ -0,0 +1,192 @@
...
+if not os.path.exists(predicts):
+    os.mkdir(predicts)
+
+
+def accuracy(a, b):
+    data = defaultdict(lambda: defaultdict(int))
+    true = 0
+    tot = 0
+    for ii, jj in zip(a, b):
+	data[ii][jj] += 1
```
vaskonov, line 37: !

requirements.txt
```diff
@@ -13,4 +13,5 @@ requests>=2.19.0,<3.0.0
 scikit-learn>=0.24,<1.1.0
 scipy<1.9.0
 tqdm>=4.42.0,<4.65.0
+pytorch-crf==0.7.2
```
vaskonov, line 16: !

deeppavlov/models/torch_bert/torch_multitask_bert.py
```diff
@@ -0,0 +1,476 @@
...
+
+    def __init__(
+        self,
+        tasks: Dict[str, Dict],
+        pretrained_bert: str = None,
+        freeze_embeddings: bool = False,
+        max_seq_len: str = 320,
+        optimizer: str = "AdamW",
+        optimizer_parameters: dict = {"lr": 2e-5},
+        config: str = "configs/top_config.json",
```
vaskonov, line 168: !

deeppavlov/models/torch_bert/torch_multitask_bert.py
```diff
@@ -0,0 +1,476 @@
...
+        for task in tasks:
+            self.task_names.append(task)
+            self.tasks_num_classes.append(tasks[task]['options'])
+            self.task_types.append(tasks[task]['type'])
+        if self.return_probas and 'sequence_labeling' in self.task_types:
+            log.warning(f'Return_probas for sequence_labeling not supported yet. Returning ids for this task')
+        self.n_tasks = len(tasks)
+        self.train_losses = [[] for task in self.task_names]
+        self.pretrained_bert = pretrained_bert
+        self.freeze_embeddings = freeze_embeddings
```
vaskonov, line 207: !

deeppavlov/configs/multitask/multitask_example.json
```diff
@@ -0,0 +1,336 @@
...
+         {
+            "id":"multitask_bert",
+            "class_name":"multitask_bert",
+            "optimizer_parameters":{
+               "lr":2e-5
+            },
+            "gradient_accumulation_steps":"{GRADIENT_ACC_STEPS}",
+            "learning_rate_drop_patience":2,
+            "learning_rate_drop_div":2.0,
+            "log_model_summary":false,
```
vaskonov, line 149: !

deeppavlov/dataset_iterators/multitask_iterator.py
```diff
@@ -13,44 +13,308 @@
...
 
-class RepeatBatchGenerator:
-    """Repeating dataset. If there is not enough elements in the dataset to form another batch, elements for the batch 
-    are drawn in the beginning of the dataset. Optionally dataset is reshuffled before a repeat.
+@register('multitask_iterator')
+class MultiTaskIterator:
+    """
+    Class merges data from several dataset iterators. When used for batch generation batches from
+    merged dataset iterators are united into one batch. If sizes of merged datasets are different
+    smaller datasets are repeated until their size becomes equal to the largest dataset.
```
vaskonov, line 35: !

PR #1606, opened 2022-12-05
New faq models

nastyachizhikova:
Update faq configs: delete not working ones, add new. Change the dataset used in the configs to CLINC150. Add few-shot setting support.


nastyachizhikova:
Opened another, clean PR

PR #1653, opened 2023-07-04
Torchcrf-import-fix

Hollyqui:
Currently the crf.py file includes the import "from torchcrf import..." which should be "from TorchCRF import..." and leads to errors when trying to load (at least) the ner_ontonotes_bert model for NER. The specific error thrown is "modulenotfounderror: No module named 'torchcrf'" despite installing the torchcrf module.


IgnatovFedor:
Hi, @Hollyqui 
Thank you for your interest in DeepPavlov Library and your PR.
I can't reproduce your error: import `from torchcrf` works like a charm, but `from TorchCRF` gives error. Could you please tell your OS, exact python version and installed DeepPavlov ant torchcrf versions?

IgnatovFedor:
upd: There is package [TorchCRF](https://pypi.org/project/TorchCRF/). I think, you installed it manually, but DeepPavlov uses [pytorch-crf](https://pypi.org/project/pytorch-crf/) package.
All model requirements could be installed either from CLI with `python -m deeppavlov install <model_name>` or from python with `model = build_model('<model_name>', install=True)`

PR #1656, opened 2023-07-11
fix: Issue 1247

### LINKED ISSUE #1247: AttributeError raised when validating with an empty validation set or a missing valid.csv
DeepPavlov 0.5.0, Python 3.7.6, Ubuntu 20.04

When the validation set is empty, this error

     AttributeError: 'NoneType' object has no attribute 'items'

is raised by

[`metrics = list(report['metrics'].items())`](https://github.com/deepmipt/DeepPavlov/blob/master/deeppavlov/core/trainers/nn_trainer.py#L174)

because [the value for `"metrics"` is set to `None` here](https://github.com/deepmipt/DeepPavlov/blob/65f69dfb898f5444cc2c98ae03ec7b3f44266df2/deeppavlov/core/trainers/fit_trainer.py#L215).

I would suggest the training to proceed with nan-values for the metrics or a clearer error to be raised when the validation set is empty but `"val_every_n_epochs"` or `"val_every_n_batches"` are set.

avglinsky:
- fix #1247

### DIFF
--- a/deeppavlov/core/commands/train.py
+++ b/deeppavlov/core/commands/train.py
@@ -92,11 +92,9 @@ def train_evaluate_model_from_config(config: Union[str, Path, dict],
     if iterator is None:
         try:
             data = read_data_by_config(config)
-            # if the value 'val_every_n_epochs' is set in the config and there is no validation data - raise exception
-            if config.get('train', {}).get('val_every_n_epochs', None) and not data.get('valid', []):
+            if config.get('train', {}).get('val_every_n_epochs') and not data.get('valid'):
                 error_message = 'The value "val_every_n_epochs" is set in the config but no validation data is provided'
-                log.exception(error_message)
-                raise Exception(error_message)
+                raise AttributeError(error_message)
         except ConfigError as e:
             to_train = False
             log.warning(f'Skipping training. {e.message}')
--- a/deeppavlov/core/commands/train.py
+++ b/deeppavlov/core/commands/train.py
@@ -92,6 +92,11 @@ def train_evaluate_model_from_config(config: Union[str, Path, dict],
     if iterator is None:
         try:
             data = read_data_by_config(config)
+            # if the value 'val_every_n_epochs' is set in the config and there is no validation data - raise exception
+            if config.get('train', {}).get('val_every_n_epochs', None) and not data.get('valid', []):
+                error_message = 'The value "val_every_n_epochs" is set in the config but no validation data is provided'
+                log.exception(error_message)
+                raise Exception(error_message)
         except ConfigError as e:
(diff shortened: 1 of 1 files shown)


deeppavlov/core/commands/train.py
```diff
@@ -92,6 +92,11 @@ def train_evaluate_model_from_config(config: Union[str, Path, dict],
     if iterator is None:
         try:
             data = read_data_by_config(config)
+            # if the value 'val_every_n_epochs' is set in the config and there is no validation data - raise exception
```
IgnatovFedor, line 95: remove comment - unnecessary

deeppavlov/core/commands/train.py
```diff
@@ -92,6 +92,11 @@ def train_evaluate_model_from_config(config: Union[str, Path, dict],
...
+            if config.get('train', {}).get('val_every_n_epochs', None) and not data.get('valid', []):
```
IgnatovFedor, line 96: add batch parameter check, replace `not data.get('valid', [])` with `not data.get('valid')`

deeppavlov/core/commands/train.py
```diff
@@ -92,6 +92,11 @@ def train_evaluate_model_from_config(config: Union[str, Path, dict],
...
+                error_message = 'The value "val_every_n_epochs" is set in the config but no validation data is provided'
+                log.exception(error_message)
```
IgnatovFedor, line 98: remove log.exception - duplicates exception output

deeppavlov/core/commands/train.py
```diff
@@ -92,6 +92,11 @@ def train_evaluate_model_from_config(config: Union[str, Path, dict],
     if iterator is None:
         try:
             data = read_data_by_config(config)
+            # if the value 'val_every_n_epochs' is set in the config and there is no validation data - raise exception
+            if config.get('train', {}).get('val_every_n_epochs', None) and not data.get('valid', []):
+                error_message = 'The value "val_every_n_epochs" is set in the config but no validation data is provided'
```
IgnatovFedor, line 97: move to exception

deeppavlov/core/commands/train.py
```diff
@@ -92,6 +92,11 @@ def train_evaluate_model_from_config(config: Union[str, Path, dict],
...
+                log.exception(error_message)
+                raise Exception(error_message)
```
IgnatovFedor, line 99: Replace with more specific error type

IgnatovFedor (changes requested):
change version in deeppavlov._metadata to 1.3.0

IgnatovFedor (approved):

PR #1665, opened 2023-10-17
fix: pydantic version

IgnatovFedor:
Fix `ImportError: cannot import name 'Undefined' from 'pydantic.fields'` on pydantic>=2


ruslan-korneev:
@IgnatovFedor maybe instead of limiting the version we could use pydantic.v1 package inside the pydantic>=2?

With your example it'll look like that, and it works fine
```python
from pydantic.v1.fields import Undefined
```

